PR #3, opened 2020-07-23
Ignore Snyk error since there's no available replacement

inespo:


tim-mwangi:
I  think you need to add `--policy-path=.snyk` to the additional arguments in in circle ci config.

PR #4, opened 2020-07-31
fix: grpc vulnerability fixed in 1.31.0

tim-mwangi:


.circleci/config.yml
```diff
@@ -89,7 +89,7 @@ jobs:
     steps:
       - setup_build_environment
       - snyk/scan:
-          additional-arguments: --all-sub-projects --policy-path=.snyk
+          additional-arguments: --all-sub-projects
```
aaron-steinfeld, line 92: May want to leave the policy path (As long as CI doesn't mind it being blank) so we don't have to modify this as ignores come and go. In fact, could we just leave the file there with no ignores to track its history easier?

tim-mwangi: ok.

tim-mwangi: done

PR #6, opened 2020-09-22
feat: add server call stream observer for rx

aaron-steinfeld:
This change adds an RxJava adapter to support converting a reactive construct (Observable, Single, Maybe, Completable - didn't add flowable support unless needed) into a GRPC response. This pattern:
- Promotes cleaner code, as an observer no longer needs to be plumbed down throw the call stack
- Supports either synchronous or async implementations (like any reactive code)
- Supports cancellation for ongoing requests

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -15,4 +15,11 @@ subprojects {
       license.set(License.APACHE_2_0)
     }
   }
+
+  pluginManager.withPlugin("java") {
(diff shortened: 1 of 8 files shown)


aaron-steinfeld:
Also bumped testing libraries, pinned to java 11 (most other repos have already done that) and updated to latest grpc.

tim-mwangi (approved):

PR #10, opened 2021-01-14
fix: unwrap thrown exceptions

aaron-steinfeld:
This changes the behavior of the execution utils to stop wrapping any thrown exceptions. This behavior was only meant to satisfy the compiler, and makes it more difficult to react based on the type of error.

### DIFF
--- a/grpc-client-rx-utils/build.gradle.kts
+++ b/grpc-client-rx-utils/build.gradle.kts
@@ -7,14 +7,9 @@ plugins {
 
 dependencies {
   api("io.reactivex.rxjava3:rxjava:3.0.6")
-  api("io.grpc:grpc-stub:1.33.1")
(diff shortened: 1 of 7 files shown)


grpc-client-rx-utils/build.gradle.kts
```diff
@@ -7,14 +7,9 @@ plugins {
 
 dependencies {
   api("io.reactivex.rxjava3:rxjava:3.0.6")
-  api("io.grpc:grpc-stub:1.33.1")
+  api("io.grpc:grpc-stub:1.35.0")
```
aaron-steinfeld, line 10: (I also took the opportunity to bump the grpc version and remove the guava constraint which is no longer needed as grpc has bumped their dependency version to 30.0-android)

tim-mwangi (approved):

PR #13, opened 2021-03-25
feat: add grpc channel registry

aaron-steinfeld:
## Description
This channel registry, ported from https://github.com/hypertrace/hypertrace-core-graphql/blob/7df21a96bf0712e239ced1117c49889870cf7b3e/hypertrace-core-graphql-grpc-utils/src/main/java/org/hypertrace/core/graphql/utils/grpc/DefaultGrpcChannelRegistry.java is used by clients to prevent multiple places in code from instantiating the same channel (which grpc strongly discourages), instead sharing channels and allowing centralized shutdown.

### Testing
Added UT

### Checklist:

### DIFF
--- a/grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
+++ b/grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
@@ -1,0 +1,34 @@
+package org.hypertrace.core.grpcutils.client;
+
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GrpcChannelRegistry {
+  private static final Logger LOG = LoggerFactory.getLogger(GrpcChannelRegistry.class);
+  private final Map<String, ManagedChannel> channelMap = new ConcurrentHashMap<>();
+  private volatile boolean isShutdown = false;
+
+  public ManagedChannel forAddress(String host, int port) {
+    assert !this.isShutdown;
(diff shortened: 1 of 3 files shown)


grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
```diff
@@ -0,0 +1,34 @@
...
+import io.grpc.ManagedChannelBuilder;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GrpcChannelRegistry {
+  private static final Logger LOG = LoggerFactory.getLogger(GrpcChannelRegistry.class);
+  private final Map<String, ManagedChannel> channelMap = new ConcurrentHashMap<>();
+  private volatile boolean isShutdown = false;
```
aaron-steinfeld, line 13: Removed the integrated shutdown on this one so it didn't bring in more dependencies (the shutdown lifecycle is from), but I can bring it back if that seems more useful. The more I think on it, the more I think I'll optionally add just the future part of it (to avoid the dependency)

skjindal93 (approved):

avinashkolluru (approved):

tim-mwangi (approved):

grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
aaron-steinfeld: Changed my mind again - the difference of accepting a hook is really just the below anyway:
```
new GrpcChannelRegistry(future);
// vs
registry = new GrpcChannelRegistry();
future.thenRun(registry::shutdown);
```

PR #15, opened 2021-05-18
feat: add jwt decoder

GurtejSohi:
## Description
Adds JWT decoder


grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/jwt/JwtParser.java
```diff
@@ -0,0 +1,94 @@
...
+    if (authHeaderValue.startsWith(BEARER_TOKEN_PREFIX)) {
+      return this.fromJwt(authHeaderValue.substring(BEARER_TOKEN_PREFIX.length()));
+    }
+
+    return Optional.empty();
+  }
+
+  public Optional<VerifiedJwt> fromJwt(String jwtValue) {
+    try {
+      return this.verifiedJwtCache.get(jwtValue, () -> this.decodeAndVerify(jwtValue));
```
surajpuvvada, line 44: why use a cache ?

surajpuvvada: if we need to access it multiple times we could store in the context object instead ?

jcchavezs:
Do not forget to include tests on this one.

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/jwt/JwtParser.java
aaron-steinfeld: So my understanding is that we actually don't need want to force verification here, instead relying on the proxy to do it. Either way though, everything around JWTs should be package protected, we only want to expose the abstracted information via the context

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/jwt/JwtParser.java
```diff
@@ -0,0 +1,94 @@
...
+      JWTVerifier verifier = this.jwtVerifierFactory.create(jwt.getKeyId());
+      verifier.verify(jwt);
+      return Optional.of(new DefaultVerifiedJwt(jwt));
+    } catch (Throwable t) {
+      LOG.warn("Failed to verify JWT", t);
+      return Optional.empty();
+    }
+  }
+
+  private static final class DefaultVerifiedJwt implements VerifiedJwt {
```
jcchavezs, line 63: I am curious about this class being final and private. Do we expect others to implement their own VerifiedJwt? if so, will that be based on this default one?

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/jwt/JwtParser.java
avinashkolluru: Yes. I think thats what we decided the other day. Verification is the responsibility of the proxy. 
Should we just add both the APIs then? Depending on the use case, the callers can decide.

aaron-steinfeld: What do you mean by both the apis? the verified and unverified one?

GurtejSohi: Removed jwt verification and now, exposing the abstracted information via the context.

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/Jwt.java
```diff
@@ -0,0 +1,13 @@
+package org.hypertrace.core.grpcutils.context;
+
+import java.util.Optional;
+
+public interface Jwt {
```
aaron-steinfeld, line 5: does this need to be public? for now at least, I think keeping the jwt completely internal to this package would be ideal.

grpc-context-utils/src/test/java/org/hypertrace/core/grpcutils/context/JwtParserTest.java
```diff
@@ -0,0 +1,58 @@
...
+  private final String testJwtUserId = "google-oauth2|108714696847779999359";
+  private final String testJwtName = "Aaron Steinfeld";
```
aaron-steinfeld, line 19: hey it's me! (but really, let's swap this out for a dummy JWT)

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/Jwt.java
GurtejSohi: Made it package-protected.

grpc-context-utils/src/test/java/org/hypertrace/core/grpcutils/context/JwtParserTest.java
GurtejSohi: Sure, replaced it with a dummy JWT

aaron-steinfeld (approved):
lgtm

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/jwt/JwtParser.java
avinashkolluru: Yes. I meant `decodeAndVerify` and just `decode`

GurtejSohi: I think we can add the verification part later on if required(we may not even need that if we're relying on proxy for verification). Also, the verification part requires passing in some extra values which makes it messy (which we may be able to avoid).

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/JwtParser.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class JwtParser {
+  private static final Logger LOG = LoggerFactory.getLogger(JwtParser.class);
+  private static final String BEARER_TOKEN_PREFIX = "Bearer ";
+
+  private final Cache<String, Optional<Jwt>> jwtCache =
```
surajpuvvada, line 17: is the cache being still used ?

GurtejSohi: Yes, using the cache to help avoid decoding the same token again.

avinashkolluru (approved):

PR #16, opened 2021-05-25
feat: add context-aware cache key

aaron-steinfeld:
## Description
In many places across many repos, we try to solve the problem of caching remote call results in different ways - usually by using the call arguments as a cache key, or sometimes the tenant id and the call arguments. This is meant to provide a reusable way of doing so, that leaves the meaningful parts of the request context as an implementation detail. The idea is that, as more data is added to our request context (for example, user identifiers), that will flow into cache keys without requiring widespread implementation changes.

### Testing
Added UTs

### Checklist:


aaron-steinfeld:
whoops - meant to draft!

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContextConstants.java
```diff
@@ -17,10 +17,20 @@
...
+
   /**
-   * The values in this set are looked up with case insensitivity.
+   * These headers may affect returned results and should be accounted for in any cached remote
+   * results
    */
-  public static final Set<String> HEADER_PREFIXES_TO_BE_PROPAGATED =
-      Set.of(TENANT_ID_HEADER_KEY, "X-B3-", "grpc-trace-bin",
-          "traceparent", "tracestate", AUTHORIZATION_HEADER);
+  static final Set<String> CACHE_MEANINGFUL_HEADERS =
```
aaron-steinfeld, line 34: I know there's some debate whether authorization is meaningful. I'll remove it if needed to unblock this since we can revisit this later in a common way, but I maintain it's better to have the default implementation err on the side of being always correct at the cost of potential duplication.

aaron-steinfeld:
@tim-mwangi @skjindal93 @avinashkolluru - can I please get a review on this? If people would prefer me remove the auth header to defer that debate that's fine, but I'd like to use this elsewhere to prevent making the ~20th implementation of this pattern :)

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/DefaultContextualKey.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+class DefaultContextualKey<T> implements ContextualKey<T> {
+  private final RequestContext context;
+  private final T data;
+  private final Map<String, String> meaningfulContextHeaders;
+
+  DefaultContextualKey(RequestContext context, T data) {
```
skjindal93, line 15: Do we always have to pass in some data to make the request context as a cache key? Can we also have a constructor, which just takes in the `RequestContext` as an argument?

skjindal93: I am a bit confused why do we need `data` (_which I believe is an input to the GRPC function calls_) as a parameter for the cache key

Basically, why should the contextual key change on a different `data`?

aaron-steinfeld: >Do we always have to pass in some data to make the request context as a cache key? Can we also have a constructor, which just takes in the RequestContext as an argument?

We don't always need to use data, there's a no arg version (this is always built off the context) https://github.com/hypertrace/java-grpc-utils/pull/16/files#diff-ef190554e9c2549e8421789de3b00e6c5fae5294934d883251699dff75676817R87 )

I'll create flavors of the function execution methods that don't provide data, too.


> I am a bit confused why do we need data (which I believe is an input to the GRPC function calls) as a parameter for the cache key
>
> Basically, why should the contextual key change on a different data?

So I attempt to convey the distinction by calling a Contextual key rather than a Context key - that is, it's a key that accounts for the context rather than a key for the context. Since the former is a superset of the latter, it felt more flexible to me as it can both handle cases where we solely want a per-context key, but also cases where we, for example, want to cache something like an entity by id.

aaron-steinfeld: Updated the first part.

avinashkolluru (approved):

skjindal93 (approved):

PR #19, opened 2021-08-23
fix: wait for termination on shutdown

aaron-steinfeld:
## Description
Trying to better manage channels so we ensure they're properly shutdown before thee JVM exits.

### Testing
Manual test

### DIFF
--- a/grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
+++ b/grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
@@ -4,6 +4,7 @@ import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -50,7 +51,37 @@ public class GrpcChannelRegistry {
   }
 
   public void shutdown() {
-    channelMap.values().forEach(ManagedChannel::shutdown);
+    channelMap.forEach(this::fullyShutdownChannel);
     this.isShutdown = true;
+  }
+
+  private void fullyShutdownChannel(String channelId, ManagedChannel managedChannel) {
+    LOG.info("Starting shutdown for channel [{}]", channelId);
+    managedChannel.shutdown();
+    if (this.waitForTermination(channelId, managedChannel)) {
+      LOG.info("Shutdown channel successfully [{}]", channelId);
+      return;
+    }
+
+    LOG.error("Shutting down channel [{}] forcefully", channelId);
+    managedChannel.shutdownNow();
+    if (this.waitForTermination(channelId, managedChannel)) {
+      LOG.info("Forced channel [{}] shutdown successful", channelId);
+    } else {
+      LOG.info("Unable to force channel [{}] shutdown - giving up!", channelId);
+    }
+  }
+
+  private boolean waitForTermination(String channelId, ManagedChannel managedChannel) {
+    try {
+      if (!managedChannel.awaitTermination(1, TimeUnit.MINUTES)) {
+        LOG.error("Channel [{}] did not shut down after 1 minute", channelId);
+      }
+    } catch (InterruptedException ex) {
+      LOG.error(
+          "There has been an interruption during the shutdown process for channel [{}]", channelId);
+      Thread.currentThread().interrupt();
+    }
+    return managedChannel.isTerminated();
   }
 }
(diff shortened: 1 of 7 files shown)


grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
```diff
@@ -50,7 +51,37 @@ private String getChannelId(String host, int port, boolean isPlaintext) {
   }
 
   public void shutdown() {
-    channelMap.values().forEach(ManagedChannel::shutdown);
+    channelMap.forEach(this::fullyShutdownChannel);
```
skjindal93, line 54: If there were N channels, it would take worst case 2*N minutes (1 min to wait for graceful shutdown + 1 min wait for forceful shutdown) for all the channels to shutdown

Should we use parallelism to shutdown channels, instead of sequential shutting down each of them, since `awaitTermination` is a thread blocking call?

aaron-steinfeld: I thought about that, but decided it felt like overkill. I'd argue this is a case where we have to look at the expected runtime vs the theoretical worst to save on complexity.

When does a shutdown hang? When there's a pending call, but we're implicitly given a longer deadline for later channels. That is, each successive channel gets the time all previous took plus another minute to complete. So the graceful part of the worst case scenario would require - N channels all with pending calls with (effectively) infinite deadlines. That's not really a scenario we have or ever should have - we always use deadlines. Now the forceful part, this should be independent of deadline since it cancels pending calls, too. My understanding is the only reason force shutdown is not instantaneous is threading - that is, the shutdown command is passed to the executor, which on complete will mark termination. So this should be quick in any scenario.

Now what happens if the worst comes to pass - every channel is in use with an infinite deadline? It should take the container N minutes to shut down, because they'd wait for 1 minute each, then forcefully shut down. So the container takes a long time to shutdown - in a k8s world, no big deal - a new replica should have been brought up and traffic diverted with this one shutting down ( we should go clean up health checks as a separate effort, btw). In a docker/compose world, it would mean it's taking a long time to shutdown and perhaps we get a little forceful - but 


Anyway, now that I've argued against it, I'll propose the following changes because there's no real harm in doing so. I still think it's overkill to bring in parallelism here, but let's call this proposal... batching :). Do you feel like the following changes would address your concerns:
- Commence all shutdowns before waiting for any
- Have a single wait period that checks all channels (we either write the wait part ourselves or share a common deadline)
- Come back through with a forceful shutdown on any unterminated channels
- Again, single (quicker - say 5s) wait for all channels.

skjindal93: > That's not really a scenario we have or ever should have - we always use deadlines

We actually don't always use deadlines (_though I know we should_)

skjindal93: > a new replica should have been brought up and traffic diverted with this one shutting down

I am guessing you mean, with this one terminating down successfully

skjindal93: I think the proposal makes sense, if it's easier to implement. Mostly, the single wait period across all channels

aaron-steinfeld: I combined the last two since it's a quick wait and force shutdown should be close to instant, but otherwise updated.

aaron-steinfeld: > We actually don't always use deadlines (_though I know we should_)

please raise issues where we don't

> I am guessing you mean, with this one terminating down successfully

Not suer I phrased it correctly but no not successfully. In that scenario, a new instance would be spun up. Once it is healthy and traffic is cut over to the new instance, the old would commence shutdown. Bad hygiene, but we don't _really_ care how long it takes to do this unless as avinash pointed out, we're using stateful sets or the like.

grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
```diff
@@ -50,7 +66,55 @@ private String getChannelId(String host, int port, boolean isPlaintext) {
   }
 
   public void shutdown() {
-    channelMap.values().forEach(ManagedChannel::shutdown);
+    this.shutdown(this.clock.instant().plus(1, MINUTES));
+  }
+
+  public void shutdown(Instant deadline) {
```
skjindal93, line 72: Can we add an explanatory comment about the entire shutdown strategy? Would be really good for a new reader, as to why did we follow this route

skjindal93 (dismissed):
This looks beautiful!

grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
aaron-steinfeld: I omitted the part about leaking and resources because I'm still not sure it's actually right, more a theory - did (hopefully) thoroughly explained how the shutdown works though.

skjindal93 (approved):

tim-mwangi (approved):

PR #21, opened 2021-08-25
refactor: work in deadlines

aaron-steinfeld:
BREAKING CHANGE: changed new apis added in last release before they're adopted

## Description
The last release hasn't been adopted yet, so I'm taking the chance to improve some of the newly added APIs. This change should only break APIs added in `0.6.0`.

The changes: remove the usage of a clock, and instead take a deadline to both shutdown channels and shutdown servers. Before, the server shutdown was relative and the channel shutdown absolute (for consistency across each channel). The deadline addresses all concerns:
- It better matches other interactions with the grpc framework
- It retains the testability since a deadline can take a test ticker, and removes the need to pass in a clock
- It's absolute, so can be shared across multiple shutdowns
- It's defined relatively, which is how we normally invoke

### Testing
Updated unit tests.

### Checklist:

### DIFF
--- a/grpc-client-utils/build.gradle.kts
+++ b/grpc-client-utils/build.gradle.kts
@@ -6,19 +6,11 @@ plugins {
 }
 
 dependencies {
+  api("io.grpc:grpc-context:1.40.0")
+  api("io.grpc:grpc-api:1.40.0")
+
   implementation(project(":grpc-context-utils"))
-
-  // Logging
   implementation("org.slf4j:slf4j-api:1.7.30")
-  // End Logging
-
-  // grpc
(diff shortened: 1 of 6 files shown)


grpc-client-utils/build.gradle.kts
```diff
@@ -6,19 +6,11 @@ plugins {
 }
 
 dependencies {
-  implementation(project(":grpc-context-utils"))
+  api("io.grpc:grpc-context:1.40.0")
```
aaron-steinfeld, line 9: also did some dependency cleanup. Could potentially have downstream affects, but only if the consumers are misconfigured.

tim-mwangi (approved):

PR #31, opened 2022-06-22
feat: add in process channel support

aaron-steinfeld:
## Description
Add a version of the channel registry that supports in process channels as well as substituting an in process channel for a known hostname.

### Testing
Added unit tests

### Checklist:

### DIFF
--- a/grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
+++ b/grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
@@ -9,13 +9,24 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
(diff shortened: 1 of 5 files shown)


grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
```diff
@@ -30,44 +41,56 @@ public ManagedChannel forSecureAddress(String host, int port) {
   }
 
   public ManagedChannel forSecureAddress(String host, int port, GrpcChannelConfig config) {
-    assert !this.isShutdown;
+    assert !this.isShutdown();
```
tim-mwangi, line 44: wow lol

aaron-steinfeld: hah - this isn't javascript, that wouldn't fly :) but the context on this change: this seemed like a reasonable thing to expose since calling a creation on a shutdown registry would lead to an error (and extending it with a subclass needed this information anyway, so rather than making it protected I properly exposed it).

PR #33, opened 2022-09-01
chore: Added spotless and explicit toString() impls

singhalprerana:
Added spotless plugin which resulted in formatting of all files..
Also, added explicit `toString()` impl for `RequestContext` and `DefaultJwt` classes


grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
```diff
@@ -138,4 +138,15 @@ public ContextualKey<Void> buildInternalContextualKey() {
   public <T> ContextualKey<T> buildInternalContextualKey(T data) {
     return new DefaultContextualKey<>(this, data, List.of(TENANT_ID_HEADER_KEY));
   }
+
+  @Override
```
aaron-steinfeld, line 142: nit: we typically use lombok so it gets updated automatically

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
```diff
@@ -138,4 +138,15 @@ public ContextualKey<Void> buildInternalContextualKey() {
...
+  public String toString() {
+    final String emptyValue = "{}";
+    return "RequestContext{"
+        + "headers="
+        + headers
+        + ", jwt="
+        + getJwt().map(Jwt::toString).orElse(emptyValue)
```
aaron-steinfeld, line 149: a bit duplicative but I suppose it makes it easier

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
singhalprerana: but doesn't lombok use the class members to generate the string? this is kind of specific to what we want - not the class members as such..

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
singhalprerana: I assumed we would want to see jwt details too

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
aaron-steinfeld: It's configurable. Can mark getter methods if you want I believe (and most of the jwt details are already exposed as getters)

aaron-steinfeld (approved):

tim-mwangi (approved):

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
singhalprerana: oh saw ur msg after merging - I guess should be ok for now...

PR #49, opened 2023-08-31
fix: for perf improvement using fast random

kotharironak:
## Description
Java's UUID.randomUUID uses SecureRandom, which internally relies on the system's /dev/random. This method requires a sufficient amount of entropy to generate random numbers and may lead to blocking.

In the present context, we are focused on generating request IDs. Therefore, a faster UUID generator based on ThreadLocalRandom should be adequate.

For a similar use case, the OpenTelemetry system also utilizes a trace ID based on ThreadRandomLocal, serving as a reference.
- https://github.com/open-telemetry/opentelemetry-java/blob/2deb6d12ef864a9191ccc5d8d353fee7daff085f/sdk/common/src/main/java/io/opentelemetry/sdk/internal/RandomSupplier.java#L27
- https://github.com/open-telemetry/opentelemetry-java/blob/2deb6d12ef864a9191ccc5d8d353fee7daff085f/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/RandomIdGenerator.java#L30

As part of this PR, using fastRadom for request-context.

### DIFF
--- a/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/FastUUIDGenerator.java
+++ b/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/FastUUIDGenerator.java
@@ -3,13 +3,13 @@ package org.hypertrace.core.grpcutils.context;
 import java.util.UUID;
 import java.util.concurrent.ThreadLocalRandom;
 
-public class UuidGenerator {
+public class FastUUIDGenerator {
   /**
    * This function generates UUIDs using ThreadLocalRandom, which is faster and doesn't block like
    * the default randomUUID method that relies on /dev/random. It's suitable for most random UUID
    * needs.
    */
-  public static UUID generateFastRandomUUID() {
+  public static UUID randomUUID() {
     long mostSigBits = ThreadLocalRandom.current().nextLong();
     long leastSigBits = ThreadLocalRandom.current().nextLong();
 
--- a/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
+++ b/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
@@ -35,7 +35,7 @@ public class RequestContext {
         .put(RequestContextConstants.TENANT_ID_HEADER_KEY, tenantId)
         .put(
             RequestContextConstants.REQUEST_ID_HEADER_KEY,
-            UuidGenerator.generateFastRandomUUID().toString());
+            FastUUIDGenerator.generateFastRandomUUID().toString());
   }
 
   public static RequestContext fromMetadata(Metadata metadata) {
--- a/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
+++ b/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
@@ -5,7 +5,6 @@ import static java.util.Objects.requireNonNull;
 import static org.hypertrace.core.grpcutils.context.RequestContextConstants.CACHE_MEANINGFUL_HEADERS;
 import static org.hypertrace.core.grpcutils.context.RequestContextConstants.TENANT_ID_HEADER_KEY;
 
-import com.github.f4b6a3.uuid.UuidCreator;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.MultimapBuilder;
@@ -36,7 +35,7 @@ public class RequestContext {
         .put(RequestContextConstants.TENANT_ID_HEADER_KEY, tenantId)
         .put(
             RequestContextConstants.REQUEST_ID_HEADER_KEY,
-            UuidCreator.getRandomBasedFast().toString());
+            UuidGenerator.generateFastRandomUUID().toString());
   }
 
   public static RequestContext fromMetadata(Metadata metadata) {
--- a/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
+++ b/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
@@ -35,7 +35,7 @@ public class RequestContext {
         .put(RequestContextConstants.TENANT_ID_HEADER_KEY, tenantId)
         .put(
             RequestContextConstants.REQUEST_ID_HEADER_KEY,
-            FastUUIDGenerator.generateFastRandomUUID().toString());
+            FastUUIDGenerator.randomUUID().toString());
   }
 
   public static RequestContext fromMetadata(Metadata metadata) {
--- a/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
+++ b/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
@@ -19,7 +19,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
@@ -35,7 +34,9 @@ public class RequestContext {
   public static RequestContext forTenantId(String tenantId) {
     return new RequestContext()
         .put(RequestContextConstants.TENANT_ID_HEADER_KEY, tenantId)
-        .put(RequestContextConstants.REQUEST_ID_HEADER_KEY, UuidCreator.getRandomBasedFast().toString());
+        .put(
+            RequestContextConstants.REQUEST_ID_HEADER_KEY,
+            UuidCreator.getRandomBasedFast().toString());
   }
 
   public static RequestContext fromMetadata(Metadata metadata) {
--- a/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
+++ b/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
@@ -5,6 +5,7 @@ import static java.util.Objects.requireNonNull;
 import static org.hypertrace.core.grpcutils.context.RequestContextConstants.CACHE_MEANINGFUL_HEADERS;
 import static org.hypertrace.core.grpcutils.context.RequestContextConstants.TENANT_ID_HEADER_KEY;
 
+import com.github.f4b6a3.uuid.UuidCreator;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.MultimapBuilder;
@@ -34,7 +35,7 @@ public class RequestContext {
   public static RequestContext forTenantId(String tenantId) {
     return new RequestContext()
         .put(RequestContextConstants.TENANT_ID_HEADER_KEY, tenantId)
-        .put(RequestContextConstants.REQUEST_ID_HEADER_KEY, UUID.randomUUID().toString());
+        .put(RequestContextConstants.REQUEST_ID_HEADER_KEY, UuidCreator.getRandomBasedFast().toString());
   }
 
   public static RequestContext fromMetadata(Metadata metadata) {
--- a/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/UuidGenerator.java
+++ b/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/UuidGenerator.java
@@ -1,0 +1,25 @@
+package org.hypertrace.core.grpcutils.context;
+
+import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
+
+public class UuidGenerator {
+  /**
+   * This function generates UUIDs using ThreadLocalRandom, which is faster and doesn't block like
+   * the default randomUUID method that relies on /dev/random. It's suitable for most random UUID
+   * needs.
+   */
+  public static UUID generateFastRandomUUID() {
+    long mostSigBits = ThreadLocalRandom.current().nextLong();
+    long leastSigBits = ThreadLocalRandom.current().nextLong();
+
+    // Set the version (4) For random UUID
+    mostSigBits &= 0xFFFFFFFFFFFF0FFFL;
+    mostSigBits |= 0x0000000000004000L;
+    // Set variant to RFC 4122
+    leastSigBits &= 0x3FFFFFFFFFFFFFFFL;
+    leastSigBits |= 0x8000000000000000L;
+
+    return new UUID(mostSigBits, leastSigBits);
+  }
+}
--- a/grpc-context-utils/build.gradle.kts
+++ b/grpc-context-utils/build.gradle.kts
@@ -17,7 +17,8 @@ dependencies {
   implementation("com.auth0:jwks-rsa:0.22.0")
   implementation("com.google.guava:guava:32.0.1-jre")
   implementation("org.slf4j:slf4j-api:1.7.36")
-
+  implementation("com.github.f4b6a3:uuid-creator:5.3.2")
+  
   annotationProcessor("org.projectlombok:lombok:1.18.24")
   compileOnly("org.projectlombok:lombok:1.18.24")
 
--- a/grpc-context-utils/build.gradle.kts
+++ b/grpc-context-utils/build.gradle.kts
@@ -18,7 +18,7 @@ dependencies {
   implementation("com.google.guava:guava:32.0.1-jre")
   implementation("org.slf4j:slf4j-api:1.7.36")
   implementation("com.github.f4b6a3:uuid-creator:5.3.2")
-  
+
   annotationProcessor("org.projectlombok:lombok:1.18.24")
   compileOnly("org.projectlombok:lombok:1.18.24")
 
--- a/grpc-context-utils/build.gradle.kts
+++ b/grpc-context-utils/build.gradle.kts
@@ -17,7 +17,6 @@ dependencies {
   implementation("com.auth0:jwks-rsa:0.22.0")
   implementation("com.google.guava:guava:32.0.1-jre")
   implementation("org.slf4j:slf4j-api:1.7.36")
-  implementation("com.github.f4b6a3:uuid-creator:5.3.2")
 
   annotationProcessor("org.projectlombok:lombok:1.18.24")
   compileOnly("org.projectlombok:lombok:1.18.24")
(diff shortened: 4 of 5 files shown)


grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/RequestContext.java
```diff
@@ -34,7 +35,7 @@ public class RequestContext {
   public static RequestContext forTenantId(String tenantId) {
     return new RequestContext()
         .put(RequestContextConstants.TENANT_ID_HEADER_KEY, tenantId)
-        .put(RequestContextConstants.REQUEST_ID_HEADER_KEY, UUID.randomUUID().toString());
+        .put(RequestContextConstants.REQUEST_ID_HEADER_KEY, UuidCreator.getRandomBasedFast().toString());
```
kotharironak, line 38: Ref: https://github.com/f4b6a3/uuid-creator/blob/3b6471642e95ca91901ce5efe817537aa0d1068b/src/main/java/com/github/f4b6a3/uuid/UuidCreator.java#L174

aaron-steinfeld: My concerns here aren't the new impl but
1. Adding a new dependency to a base lib like this impacts the dependency tree of every single service
2. This is just covering up an upstream problem - that we're generating request contexts inappropriately - many times for a single "request".

kotharironak: >  Adding a new dependency to a base lib like this impacts the dependency tree of every single service

I can extract out the function from this library based on ThreadLocalRandom.

>This is just covering up an upstream problem - that we're generating request contexts inappropriately - many times for a single "request".

I didn't fully get this. Can you elaborate?

aaron-steinfeld: > I didn't fully get this. Can you elaborate?

The reason this came up - at least the reason I'm aware of - as an issue is because ingester is generating many request contexts for a single trace ingestion. Each time it needs a context, it generates a new one. Because this now occurs across many enrichers and many threads in parallel, we're running into these performance problems. But not only is that unnecessary, it's actually wrong - we want the same ID to be shared for ingesting the same trace in the first place.

A couple different approaches there would be to either
1. update the enricher interface to accept the context, and generate it up front before we delegate the enrichment off to an async thread.
2. Create a request context from the trace details - use the trace id as the "request id". This would probably be defined in the ingestion pipeline repo.

mihirgt: The request-id attribute is attached to the grpc-context while making a downstream call. This also seems to be getting added only for RXJava calls, if someone is not using RXJava, this request-id is not attached to the call.
How is this request-id used by the downstream services? Given that its attached only for RXJava based calls and not for all GRPC calls, do we even need to attach this request-id?

kotharironak: >The reason this came up - at least the reason I'm aware of - as an issue is because ingester is generating many request contexts for a single trace ingestion. Each time it needs a context, it generates a new one. Because this now occurs across many enrichers and many threads in parallel, we're running into these performance problems. But not only is that unnecessary, it's actually wrong - we want the same ID to be shared for ingesting the same trace in the first place.

> A couple different approaches there would be to either

> update the enricher interface to accept the context, and generate it up front before we delegate the enrichment off to an async thread.
Create a request context from the trace details - use the trace id as the "request id". This would probably be defined in the ingestion pipeline repo.

I agree that for tracing the processing of StructureTrace, it's important to establish a context based on unique keys or attributes that can be used to generate a trace ID. And if `request_id` is primarily used internally to link outgoing requests to the same trace, we can consider option (2) having new request context based trace.

However, for any other consumers of this interface `forTenantId`, do we really need a `request_id` generated from a secure random source? Since this is primarily for tracing purposes and not for resource identification, I think it's more practical to use a fast UUID-based `request_id` generated with `ThreadLocalRandom`.

kotharironak: > I can extract out the function from this library based on ThreadLocalRandom.

If we don't plan to use the above library here. I am thinking of having our own function like below:
```
public static UUID generateFastRandomUUID() {
    long mostSigBits = ThreadLocalRandom.current().nextLong();
    long leastSigBits = ThreadLocalRandom.current().nextLong();
    
    // Set the version (4) For random UUID 
    mostSigBits &= 0xFFFFFFFFFFFF0FFFL;
    mostSigBits |= 0x0000000000004000L;
    // Set variant to RFC 4122
    leastSigBits &= 0x3FFFFFFFFFFFFFFFL;  
    leastSigBits |= 0x8000000000000000L; 

    return new UUID(mostSigBits, leastSigBits);
}
```

laxmanchekka: Few stacks showing up the contention at jdk level

```
"kafka-streams-async-worker-3" #97 daemon prio=5 os_prio=0 cpu=158592.16ms elapsed=573.69s tid=0x00007fdd9b4fb800 nid=0x570 waiting for monitor entry  [0x00007fdd6f422000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at sun.security.provider.NativePRNG$RandomIO.implNextBytes(java.base@11.0.19/Unknown Source)
	- waiting to lock <0x0000000115d07748> (a java.lang.Object)
	at sun.security.provider.NativePRNG.engineNextBytes(java.base@11.0.19/Unknown Source)
	at java.security.SecureRandom.nextBytes(java.base@11.0.19/Unknown Source)
	at java.util.UUID.randomUUID(java.base@11.0.19/Unknown Source)
	at org.hypertrace.core.grpcutils.context.RequestContext.forTenantId(RequestContext.java:37)
	at ai.traceable.platform.enrichment.clients.ApiAttributeOverrideCacheWithInvalidation.getApiIsExternalOverride(ApiAttributeOverrideCacheWithInvalidation.java:54)
	at ai.traceable.platform.apientity.ApiEntityDataAccessManager.evaluateIsExternalApi(ApiEntityDataAccessManager.java:499)
	at ai.traceable.platform.apientity.ApiEntityDataAccessManager.getUpdatedAttributes(ApiEntityDataAccessManager.java:253)
	at ai.traceable.platform.apientity.AbstractApiDataAccessManager.updateApiEntity(AbstractApiDataAccessManager.java:196)
	at ai.traceable.platform.apientity.http.discovery.HttpApiDiscoveryManager.getEntityForPattern(HttpApiDiscoveryManager.java:184)
	at ai.traceable.platform.apientity.http.discovery.HttpApiDiscoveryManager.getApiEntities(HttpApiDiscoveryManager.java:126)
	at ai.traceable.platform.enrichment.enrichers.ApiAttributeEnricher.enrichEvent(ApiAttributeEnricher.java:231)
	at org.hypertrace.traceenricher.enrichment.EnrichmentProcessor.applyEnricher(EnrichmentProcessor.java:123)
	at org.hypertrace.traceenricher.enrichment.EnrichmentProcessor.process(EnrichmentProcessor.java:81)
	at ai.traceable.platform.enrichment.enrichers.TraceableConvertingEnrichmentProcessor.enrichAndConvert(TraceableConvertingEnrichmentProcessor.java:38)
	at ai.traceable.platform.trace.enricher.StructuredTraceEnrichProcessor.asyncProcess(StructuredTraceEnrichProcessor.java:75)
	at ai.traceable.platform.trace.enricher.StructuredTraceEnrichProcessor.asyncProcess(StructuredTraceEnrichProcessor.java:26)
	at org.hypertrace.core.kafkastreams.framework.async.AsyncProcessor.lambda$process$0(AsyncProcessor.java:64)
	at org.hypertrace.core.kafkastreams.framework.async.AsyncProcessor$$Lambda$1077/0x00000008007df040.get(Unknown Source)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(java.base@11.0.19/Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(java.base@11.0.19/Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(java.base@11.0.19/Unknown Source)
	at java.lang.Thread.run(java.base@11.0.19/Unknown Source)
```


```
"kafka-streams-async-worker-15" #109 daemon prio=5 os_prio=0 cpu=162114.71ms elapsed=585.83s tid=0x00007fdd9b560000 nid=0x57c waiting for monitor entry  [0x00007fdd6dc16000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at sun.security.provider.NativePRNG$RandomIO.implNextBytes(java.base@11.0.19/Unknown Source)
	- waiting to lock <0x0000000115d07748> (a java.lang.Object)
	at sun.security.provider.NativePRNG.engineNextBytes(java.base@11.0.19/Unknown Source)
	at java.security.SecureRandom.nextBytes(java.base@11.0.19/Unknown Source)
	at java.util.UUID.randomUUID(java.base@11.0.19/Unknown Source)
	at org.hypertrace.core.grpcutils.context.RequestContext.forTenantId(RequestContext.java:37)
	at org.hypertrace.core.grpcutils.client.rx.GrpcRxExecutionContext.forTenantContext(GrpcRxExecutionContext.java:47)
	at org.hypertrace.trace.accessor.entities.DefaultTraceEntityAccessor.spanTenantContext(DefaultTraceEntityAccessor.java:197)
	at org.hypertrace.trace.accessor.entities.DefaultTraceEntityAccessor.resolveAllAttributes(DefaultTraceEntityAccessor.java:168)
	at org.hypertrace.trace.accessor.entities.DefaultTraceEntityAccessor.buildEntity(DefaultTraceEntityAccessor.java:125)
	at org.hypertrace.trace.accessor.entities.DefaultTraceEntityAccessor.writeEntityIfExists(DefaultTraceEntityAccessor.java:65)
	at org.hypertrace.trace.accessor.entities.DefaultTraceEntityAccessor.lambda$writeAssociatedEntitiesForSpanEventually$1(DefaultTraceEntityAccessor.java:61)
	at org.hypertrace.trace.accessor.entities.DefaultTraceEntityAccessor$$Lambda$1518/0x0000000800b3f440.accept(Unknown Source)
	at java.util.ArrayList.forEach(java.base@11.0.19/Unknown Source)
	at org.hypertrace.trace.accessor.entities.DefaultTraceEntityAccessor.writeAssociatedEntitiesForSpanEventually(DefaultTraceEntityAccessor.java:61)
	at org.hypertrace.traceenricher.enrichment.enrichers.EntitySpanEnricher.enrichEvent(EntitySpanEnricher.java:19)
	at ai.traceable.platform.enrichment.enrichers.FeatureFlaggedEntitySpanEnricher.lambda$enrichEvent$2(FeatureFlaggedEntitySpanEnricher.java:35)
	at ai.traceable.platform.enrichment.enrichers.FeatureFlaggedEntitySpanEnricher$$Lambda$1510/0x0000000800b3cc40.accept(Unknown Source)
	at io.reactivex.rxjava3.internal.observers.BlockingMultiObserver.blockingConsume(BlockingMultiObserver.java:175)
	at io.reactivex.rxjava3.core.Maybe.blockingSubscribe(Maybe.java:2990)
	at io.reactivex.rxjava3.core.Maybe.blockingSubscribe(Maybe.java:2935)
	at ai.traceable.platform.enrichment.enrichers.FeatureFlaggedEntitySpanEnricher.enrichEvent(FeatureFlaggedEntitySpanEnricher.java:35)
	at org.hypertrace.traceenricher.enrichment.EnrichmentProcessor.applyEnricher(EnrichmentProcessor.java:123)
	at org.hypertrace.traceenricher.enrichment.EnrichmentProcessor.process(EnrichmentProcessor.java:81)
	at ai.traceable.platform.enrichment.enrichers.TraceableConvertingEnrichmentProcessor.enrichAndConvert(TraceableConvertingEnrichmentProcessor.java:38)
	at ai.traceable.platform.trace.enricher.StructuredTraceEnrichProcessor.asyncProcess(StructuredTraceEnrichProcessor.java:75)
	at ai.traceable.platform.trace.enricher.StructuredTraceEnrichProcessor.asyncProcess(StructuredTraceEnrichProcessor.java:26)
```

aaron-steinfeld:
> The request-id attribute is attached to the grpc-context while making a downstream call. This also seems to be getting added only for RXJava calls, if someone is not using RXJava, this request-id is not attached to the call.
How is this request-id used by the downstream services? Given that its attached only for RXJava based calls and not for all GRPC calls, do we even need to attach this request-id?

This is unrelated to rxjava. RequestContext and everything it contains (tenant id, request id, auth context where applicable) is used by all of our grpc calls (a handful of which are themselves managed by rxjava). This request ID is used to correlate errors and logs across services.


> I agree that for tracing the processing of StructureTrace, it's important to establish a context based on unique keys or attributes that can be used to generate a trace ID. And if request_id is primarily used internally to link outgoing requests to the same trace, we can consider option (2) having new request context based trace.

The unique key is just the trace id itself though, right? Either way I think it would make sense to update the enricher interface so however we generate the request context is immaterial to the enrichers, it's just a question of level of effort. If that's painful, then having a locally consistent version that does something like `buildRequestContextForTrace(structuredTrace)` seems reasonable, since that's what we should be doing at the beginning of the enricher regardless.

> However, for any other consumers of this interface forTenantId, do we really need a request_id generated from a secure random source? Since this is primarily for tracing purposes and not for resource identification, I think it's more practical to use a fast UUID-based request_id generated with ThreadLocalRandom.

So here is where I think we may be getting too far in the weeds. I agree, it doesn't need to be secure and am certainly glad to have a thread local implementation (yours looks fine) - but I think it becomes an unnecessary optimization if we fix the request context generation itself. We're fixing a symptom of an upstream bug rather than the bug itself.


> Few stacks showing up the contention at jdk level

Right, this is the one coming from the context generation being inside the async threads rather than up front. Edited to remove internal details.

laxmanchekka:
> if we fix the request context generation itself
GRPC request context generation and implementation is the responsibility of the grpc framework itself. Why do we want to push this responsibility down to the applications and leak/duplicate this code everywhere.

While I agree that trace enricher refactoring is needed for several other good reasons. Definitely not worth for this one.

laxmanchekka:
@kotharironak : I vote for a simple util based implementation without third party dependency.

aaron-steinfeld:
> GRPC request context generation and implementation is the responsibility of the grpc framework itself. Why do we want to push this responsibility down to the grpc client applications and leak/duplicate this code everywhere.

I would still leave it there, it's just a new constructor allowing callers to bring their own ID when that makes sense.

```java
  public static RequestContext forTenantId(String tenantId) {
    return RequestContext.forTenantAndRequestIds(tenantId, UUID.randomUUID().toString()); // This would be the added api
  }
```
There's no real duplicate code, as this only impacts clients in that they're adding the extra arg from data they already have where applicable. And that's unknowable data to the framework.

> While I agree that trace enricher refactoring is needed for several other good reasons. Definitely not worth for this one.

We can choose how major of a refactor we want this to be. The interface can be left as is or updated, but we can't fix the actual bug here without changing the duplicated `RequestContext.forTenantId` calls. We could fix the thread contention with the secure random change, but not the broken correlation. IMO, the minimum/maximum effort here isn't really that significant - on the scale of hours.

mihirgt:
Given that UUID.randomUUID takes a lock and causes thread contention, we should replace it with a a ThreadLocalRandom based implementation. This will help in general. 
In follow up PRs we can fix the callers to fix calling context in trace enricher.

aaron-steinfeld:
> Given that UUID.randomUUID takes a lock and causes thread contention, we should replace it with a a ThreadLocalRandom based implementation. This will help in general. In follow up PRs we can fix the callers to fix calling context in trace enricher.

I just don't want us to lose sight of the root cause here, but otherwise no issues with that.

kotharironak:
> > Given that UUID.randomUUID takes a lock and causes thread contention, we should replace it with a a ThreadLocalRandom based implementation. This will help in general. In follow up PRs we can fix the callers to fix calling context in trace enricher.
> 
> I just don't want us to lose sight of the root cause here, but otherwise no issues with that.

@laxmanchekka @mihirgt @aaron-steinfeld 
I have added the static function, and removes the usage of third-party lib.
I will work on separate PR for context change for enricher.

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/UuidGenerator.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
+
+public class UuidGenerator {
+  /**
+   * This function generates UUIDs using ThreadLocalRandom, which is faster and doesn't block like
+   * the default randomUUID method that relies on /dev/random. It's suitable for most random UUID
+   * needs.
+   */
+  public static UUID generateFastRandomUUID() {
```
laxmanchekka, line 12:
```suggestion
  public static UUID randomUUID() {
```

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/UuidGenerator.java
```diff
@@ -0,0 +1,25 @@
+package org.hypertrace.core.grpcutils.context;
+
+import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
+
+public class UuidGenerator {
```
laxmanchekka, line 6:
```suggestion
public class FastUUIDGenerator {
```

laxmanchekka (approved):

laxmanchekka (approved):

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/FastUUIDGenerator.java
```diff
@@ -0,0 +1,25 @@
+package org.hypertrace.core.grpcutils.context;
+
+import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
+
+public class FastUUIDGenerator {
```
aaron-steinfeld, line 6: Does this need to be public? Doesn't need a new release, but next time you're in here can you move this to package private since it's an internal impl detail.

kotharironak: Initially thought of it to our own lib so it can be used. But, I think, we can make it private, anywhere else we can use the third-party library that is already there in most of the projects.

kotharironak: @laxmanchekka @aaron-steinfeld https://github.com/hypertrace/java-grpc-utils/pull/50/files

PR #55, opened 2023-12-15
feat: standard exceptions

aaron-steinfeld:
## Description

Add proposal for standard exceptions. Included tests that show how exceptions would be built.
Also:
- Server exception interceptor now sends request context back if trailers aren't already explicitly included
- RequestContext updated to include tenant id in trailers
- RequestContext now supports equals/hashcode
- RequestContext JWT Parser is now static and shared. This seems to match the original intent, as the JWT Parser includes a cache that would only make sense if shared across context. Should help avoid allocations and reduce JWT parsing.

Added UTs. Expect this to be fully backwards compatible.

### DIFF
--- a/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/ContextualStatusExceptionBuilder.java
+++ b/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/ContextualStatusExceptionBuilder.java
@@ -1,0 +1,37 @@
+package org.hypertrace.core.grpcutils.context;
+
+import io.grpc.Status;
+import io.grpc.StatusException;
+import io.grpc.StatusRuntimeException;
+import javax.annotation.Nonnull;
+import lombok.AccessLevel;
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class ContextualStatusExceptionBuilder {
+
+  private final Status status;
+  ContextualExceptionDetails details;
+
+  public static ContextualStatusExceptionBuilder from(Status status, RequestContext context) {
+    return new ContextualStatusExceptionBuilder(status, new ContextualExceptionDetails(context));
+  }
+
+  public ContextualStatusExceptionBuilder useStatusDescriptionAsExternalMessage() {
+    this.details = this.details.withExternalMessage(this.status.getDescription());
+    return this;
+  }
+
+  public ContextualStatusExceptionBuilder withExternalMessage(@Nonnull String externalMessage) {
+    this.details = this.details.withExternalMessage(externalMessage);
+    return this;
+  }
+
+  public StatusRuntimeException buildRuntimeException() {
+    return status.asRuntimeException(this.details.toMetadata());
+  }
+
+  public StatusException buildCheckedException() {
+    return status.asException(this.details.toMetadata());
+  }
+}
--- a/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/ContextualStatusExceptionBuilder.java
+++ b/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/ContextualStatusExceptionBuilder.java
@@ -6,7 +6,13 @@ import io.grpc.StatusRuntimeException;
 import javax.annotation.Nonnull;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
 
+@Getter
+@EqualsAndHashCode
+@ToString
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 public class ContextualStatusExceptionBuilder {
 
--- a/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/ContextualExceptionDetails.java
+++ b/grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/ContextualExceptionDetails.java
@@ -17,7 +17,7 @@ public class ContextualExceptionDetails {
   @Nonnull RequestContext requestContext;
   @Nullable String externalMessage;
 
-  Optional<String> getExternalMessage() {
+  public Optional<String> getExternalMessage() {
     return Optional.ofNullable(this.externalMessage);
   }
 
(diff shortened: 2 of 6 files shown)


tim-mwangi (approved):

grpc-context-utils/src/main/java/org/hypertrace/core/grpcutils/context/ContextualStatusExceptionBuilder.java
```diff
@@ -0,0 +1,43 @@
...
+import lombok.Getter;
+import lombok.ToString;
+
+@Getter
+@EqualsAndHashCode
+@ToString
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class ContextualStatusExceptionBuilder {
+
+  private final Status status;
```
singhalprerana, line 19: I'm super confused by this design.. builder has `status` but actual exception class has request context? why do we need these 2 classes anyways? can't we keep it simple and just have the one class `ContextualStatusException` consisting of all requestContext, status and externalMessage? then there can be methods to return whatever needed - statusRuntimeException or externalMessage..

aaron-steinfeld: > I'm super confused by this design.. builder has `status` but actual exception class has request context? why do we need these 2 classes anyways? can't we keep it simple and just have the one class `ContextualStatusException` consisting of all requestContext, status and externalMessage? then there can be methods to return whatever needed - statusRuntimeException or externalMessage..

Remember from our earlier discussion - we're not creating our own exception class since the grpc framework already works with status exceptions. So we've created a payload for it (the detail class) and a builder to assemble an exception with the payload. The builder needs the details of the status, context and payload to that, but produces a single exception.

In other words, the details class is an impl detail from the throwers perspective. It's never seen directly. On the receiving side, it's the other way around. The details deserialize but not the builder.

singhalprerana:
```
@Value
@AllArgsConstructor(access = AccessLevel.PACKAGE)
public class ContextualException {
  static final Metadata.Key<String> EXTERNAL_MESSAGE_KEY =
      Metadata.Key.of("external-message", Metadata.ASCII_STRING_MARSHALLER);

  @Nonnull RequestContext requestContext;
  @Nonnull Status status;
  @Nullable String externalMessage;

  public static ContextualException from(RequestContext requestContext, Status status) {
    return new ContextualException(requestContext, status, null);
  }

  public static ContextualException from(
      RequestContext requestContext, Status status, String externalMessage) {
    return new ContextualException(requestContext, status, externalMessage);
  }

  public ContextualException exceptionWithStatusDescriptionAsExternalMessage() {
    return new ContextualException(
        this.getRequestContext(), this.getStatus(), this.getStatus().getDescription());
  }

  public Optional<String> getExternalMessage() {
    return Optional.ofNullable(this.externalMessage);
  }

  public StatusRuntimeException asStatusRuntimeException() {
    return status.asRuntimeException(toMetadata());
  }

  public StatusException asStatusException() {
    return status.asException(toMetadata());
  }

  private Metadata toMetadata() {
    Metadata metadata = new Metadata();
    this.getExternalMessage().ifPresent(value -> metadata.put(EXTERNAL_MESSAGE_KEY, value));
    metadata.merge(this.getRequestContext().buildTrailers());
    return metadata;
  }
}
```
something like this - wouldn't it be simpler to understand and use.?

singhalprerana: Also, shouldn't there be a static method to extract externalMessage from request context metadata?

singhalprerana: > we're not creating our own exception class since the grpc framework already works with status exceptions

Agreed (we can rename the class in my above code snippet) but still the bifurcation seems confusing to me.. We can have a single class which owns the responsibility of both constructing and extracting the externalMessage from the metadata..

aaron-steinfeld: > something like this - wouldn't it be simpler to understand and use?

I'm not sure I'm getting the improvement.
1. This class is called "Exception", but isn't actually an exception - you've basically just combined the builder and payload into one. Edit: just saw the comment about renaming.
2. This version doesn't handle deserialization.

> Also, shouldn't there be a static method to extract externalMessage from request context metadata?

There's a static method to extract the exception details from the trailers - since that's the information you'll have at deserialization time. `ContextualExceptionDetails.fromMetadata`

> Agreed (we can rename the class in my above code snippet) but still the bifurcation seems confusing to me.. We can have a single class which owns the responsibility of both constructing and extracting the externalMessage from the metadata..

I think this is the crux of the disconnect. I'm intentionally separating the responsibilities. The exception producing side is only ever using `ContextualStatusExceptionBuilder` - the `Details` class represents the built output. The builder's  responsibility is for producing an exception with a payload out of all the various inputs. 

The exception receiving side is only ever using `ContextualExceptionDetails` - this class represents the exception payload. It would never have to know any details about the message was built.

singhalprerana: Ok so the producer of the exception is going to use this class something like `throw ContextualStatusExceptionBuilder.from().buildRuntimeException()`
and consumer would use `ContextualExceptionDetails.fromMetadata().map(Details::getExternalMessage)` to get the externalMessage?

aaron-steinfeld: Correct. 

Producing exception:
```java
   throw ContextualStatusExceptionBuilder.from(
                Status.INVALID_ARGUMENT.withDescription("Some description"), TEST_CONTEXT)
            .useStatusDescriptionAsExternalMessage()
            .buildRuntimeException();
```

Reading from exception of unknown type:
```java
      Status status = Status.fromThrowable(throwable);
      ContextualExceptionDetails details = ContextualExceptionDetails.fromMetadata(Status.trailersFromThrowable(throwable));
```

Reading from a known status exception:
``` java
      Status status = exception.getStatus();
      ContextualExceptionDetails details = ContextualExceptionDetails.fromMetadata(exception.getTrailers());
```

singhalprerana (approved):

PR #56, opened 2023-12-29
refactor: add support for better exception propagation and creation

aaron-steinfeld:
## Description

Made context propagation automatic, removed requirement to provide context when creating context.

### DIFF
--- a/grpc-server-utils/src/main/java/org/hypertrace/core/grpcutils/server/ThrowableResponseInterceptor.java
+++ b/grpc-server-utils/src/main/java/org/hypertrace/core/grpcutils/server/ThrowableResponseInterceptor.java
@@ -1,11 +1,15 @@
 package org.hypertrace.core.grpcutils.server;
 
+import static io.grpc.Status.Code.UNKNOWN;
+import static java.util.Objects.isNull;
+
 import io.grpc.ForwardingServerCall;
 import io.grpc.Metadata;
 import io.grpc.ServerCall;
 import io.grpc.ServerCallHandler;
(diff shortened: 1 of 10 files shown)


grpc-server-utils/src/main/java/org/hypertrace/core/grpcutils/server/ThrowableResponseInterceptor.java
```diff
@@ -22,21 +26,49 @@ public void sendMessage(RespT message) {
...
+    if (isNull(cause)) {
+      return new Metadata();
+    }
+    Status statusFromCause = Status.fromThrowable(cause);
+    if (statusFromCause.getCode() == UNKNOWN) {
+      return new Metadata();
+    }
+    // Otherwise, we've found a status so collect any trailers from it and merge them on top of
+    // any trailers we can find from descendents
+    Metadata trailersFromCauseDescendents = this.collectAllTrailersFromCause(statusFromCause);
```
tim-mwangi, line 69: Could this collection grow out of bounds affecting memory?

aaron-steinfeld: Not really a concern. It only recurses for the length of the causal chain, so it's bounded by the size of that. As soon as we run out of causes or hit an unknown cause, we exit recursion (the two base cases above). The causal chain must necessarily be the same depth as this recursion (or longer), and each link in that is a throwable that contains a superset of the data we're copying out.

tim-mwangi (approved):

PR #58, opened 2024-04-17
Add additional log to track methods gettign RESOURCE_EXHAUSTED error

mihirgt:


grpc-server-utils/src/main/java/org/hypertrace/core/grpcutils/server/ThrowableResponseInterceptor.java
```diff
@@ -34,6 +39,14 @@ public void close(Status status, Metadata trailers) {
                       .withDescription(status.getCause().getMessage())
                       .withCause(status.getCause());
             }
+
+            // Logging this for RESOURCE_EXHAUSTED so that we can track which method is being called
```
aaron-steinfeld, line 43: We already have metrics here, why do we need to log?

mihirgt: @aaron-steinfeld you approve this? The metrics are not useful in tracking the calling service.

aaron-steinfeld: I don't particularly like special casing this - the metrics show up on the client side as I was showing you when we discussed earlier. The one that was causing the issue you saw was missing but only because the caller was not using the utils in the first place

PR #61, opened 2024-05-31
chore: add external error interceptor

aaron-steinfeld:
## Description

This interceptor is meant to be used in edge services to prevent leaking any sensitive information. It only propagates status code, an external message and request ID, if set. All other metadata such as cause and metadata are stripped.

### DIFF
--- a/grpc-server-utils/src/main/java/org/hypertrace/core/grpcutils/server/ExternalExceptionInterceptor.java
+++ b/grpc-server-utils/src/main/java/org/hypertrace/core/grpcutils/server/ExternalExceptionInterceptor.java
@@ -63,7 +63,7 @@ public class ExternalExceptionInterceptor implements ServerInterceptor {
       return Status.fromCode(status.getCode())
           .withDescription(
               String.format("Request with id: %s failed with message: %s", requestId, message));
-    }c
+    }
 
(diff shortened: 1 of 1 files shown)


sarthak77 (dismissed):

grpc-server-utils/src/main/java/org/hypertrace/core/grpcutils/server/ExternalExceptionInterceptor.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.UUID;
+import org.hypertrace.core.grpcutils.context.ContextualExceptionDetails;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+import org.hypertrace.core.grpcutils.context.RequestContextConstants;
+
+/**
+ * This interceptor can be used at the edge to scrub any sensitive information such as an exception
+ * cause or metadata off the context before propagating it
+ */
+public class ExternalExceptionInterceptor implements ServerInterceptor {
```
sarthak77, line 23: Should we define this as an abstract class so that one can define custom message and custom trailer building logic 🤔

tim-mwangi (approved):

PR #63, opened 2024-09-24
fix: restore netty constraint

aaron-steinfeld:
Restore netty constraint

### DIFF
--- a/grpc-client-utils/build.gradle.kts
+++ b/grpc-client-utils/build.gradle.kts
@@ -11,6 +11,9 @@ dependencies {
   api("io.grpc:grpc-context")
   api("io.grpc:grpc-api")
   api("io.grpc:grpc-inprocess")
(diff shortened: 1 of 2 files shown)


grpc-client-utils/build.gradle.kts
```diff
@@ -11,6 +11,9 @@ dependencies {
   api("io.grpc:grpc-context")
   api("io.grpc:grpc-api")
   api("io.grpc:grpc-inprocess")
+  api(platform("io.netty:netty-bom:4.1.108.Final")) {
```
aaron-steinfeld, line 14: My mistake - I thought this was an old constraint since it wasn't being flagged on this repo and removed it. But it's actually still used upstream when consuming grpc (where we actually bring in netty).

tim-mwangi (approved):

PR #68, opened 2025-03-11
Add grpc circuit breaker utility using interceptors

pavan-traceable:
## Description
Added grpc utility to support circuit breaker for slow and failure calls 

### Testing
Tested changes on sandbox environment using this utility.


grpc-circuitbreaker-utils/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+  implementation("io.grpc:grpc-core")
+  implementation("io.github.resilience4j:resilience4j-circuitbreaker:1.7.1")
+  implementation("com.typesafe:config:1.4.2")
+  implementation("com.google.inject:guice:7.0.0")
+  implementation("org.hypertrace.core.serviceframework:platform-metrics:0.1.87")
+
+  annotationProcessor("org.projectlombok:lombok:1.18.24")
+  compileOnly("org.projectlombok:lombok:1.18.24")
+
+  testImplementation("org.junit.jupiter:junit-jupiter:5.8.2")
```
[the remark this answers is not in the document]
pavan-traceable: done

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfigProvider.java
```diff
@@ -0,0 +1,112 @@
...
+import io.github.resilience4j.circuitbreaker.CircuitBreakerConfig;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class CircuitBreakerConfigProvider {
+
+  public static final String CIRCUIT_BREAKER_CONFIG = "circuit.breaker.config";
```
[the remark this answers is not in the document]
pavan-traceable: Fixed it, please check now

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfigProvider.java
```diff
@@ -0,0 +1,112 @@
...
+  // Minimum calls before evaluating failure rate
+  private static final String MINIMUM_NUMBER_OF_CALLS = "minimumNumberOfCalls";
+  // Calls allowed in HALF_OPEN state before deciding to
+  // CLOSE or OPEN again
+  private static final String PERMITTED_NUMBER_OF_CALLS_IN_HALF_OPEN_STATE =
+      "permittedNumberOfCallsInHalfOpenState";
+  private static final String SLIDING_WINDOW_TYPE = "slidingWindowType";
+
+  // Cache for storing CircuitBreakerConfig instances
+  private static final ConcurrentHashMap<String, CircuitBreakerConfig> configCache =
```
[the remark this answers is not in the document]
pavan-traceable: done

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerModule.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @Provides
+  @Singleton
+  public CircuitBreakerConfigProvider providesCircuitBreakerConfigProvider() {
+    return new CircuitBreakerConfigProvider(config);
+  }
+
+  @Provides
+  @Singleton
+  public CircuitBreakerRegistry providesCircuitBreakerRegistry(
```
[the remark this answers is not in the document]
pavan-traceable: The CircuitBreakerRegistry is a factory to create CircuitBreaker instances which stores all CircuitBreaker instances in a registry. This is from https://github.com/resilience4j/resilience4j/blob/master/resilience4j-circuitbreaker/src/main/java/io/github/resilience4j/circuitbreaker/CircuitBreakerRegistry.java

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerModule.java
```diff
@@ -0,0 +1,67 @@
...
+            entryRemovedEvent -> {
+              CircuitBreaker removedCircuitBreaker = entryRemovedEvent.getRemovedEntry();
+              log.info("CircuitBreaker {} removed", removedCircuitBreaker.getName());
+            });
+    return circuitBreakerRegistry;
+  }
+
+  @Provides
+  @Singleton
+  public CircuitBreakerInterceptor providesCircuitBreakerInterceptor(
```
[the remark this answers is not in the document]
pavan-traceable: Removed it now. please check

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerEventListener.java
```diff
@@ -0,0 +1,38 @@
...
+import io.github.resilience4j.circuitbreaker.CircuitBreaker;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class CircuitBreakerEventListener {
+  private static final Set<String> attachedCircuitBreakers = ConcurrentHashMap.newKeySet();
+
+  public static synchronized void attachListeners(CircuitBreaker circuitBreaker) {
```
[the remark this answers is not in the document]
pavan-traceable: Fixed it

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerInterceptor.java
```diff
@@ -0,0 +1,130 @@
...
+  // Intercepts the call and applies circuit breaker logic
+  @Override
+  public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+      MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+    if (!circuitBreakerConfigProvider.isCircuitBreakerEnabled()) {
+      return next.newCall(method, callOptions);
+    }
+
+    // Get circuit breaker key from CallOptions
+    String circuitBreakerKey = callOptions.getOption(CIRCUIT_BREAKER_KEY);
```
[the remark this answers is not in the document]
pavan-traceable: I didn't understand this, did lot of refactoring though now.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerInterceptor.java
```diff
@@ -0,0 +1,130 @@
...
+    CircuitBreaker circuitBreaker = getCircuitBreaker(circuitBreakerKey);
+    return new ForwardingClientCall.SimpleForwardingClientCall<>(
+        next.newCall(method, callOptions)) {
+      @Override
+      public void start(Listener<RespT> responseListener, Metadata headers) {
+        long startTime = System.nanoTime();
```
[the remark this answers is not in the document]
pavan-traceable: done

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerInterceptor.java
```diff
@@ -0,0 +1,130 @@
...
+
+      @Override
+      public void sendMessage(ReqT message) {
+        if (!circuitBreaker.tryAcquirePermission()) {
+          handleCircuitBreakerRejection(circuitBreakerKey, circuitBreaker);
+          String rejectionReason =
+              circuitBreaker.getState() == CircuitBreaker.State.HALF_OPEN
+                  ? "Circuit Breaker is HALF-OPEN and rejecting excess requests"
+                  : "Circuit Breaker is OPEN and blocking requests";
+          throw Status.UNAVAILABLE.withDescription(rejectionReason).asRuntimeException();
```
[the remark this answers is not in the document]
pavan-traceable: done

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerInterceptor.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+    return circuitBreakerKey.split("\\.", 2)[0]; // Ensures only the first split
+  }
+
+  /** Retrieve the Circuit Breaker based on the key. */
+  private CircuitBreaker getCircuitBreaker(String circuitBreakerKey) {
+    CircuitBreaker circuitBreaker =
+        circuitBreakerRegistry.circuitBreaker(
+            circuitBreakerKey, circuitBreakerConfigProvider.getConfig(circuitBreakerKey));
+    CircuitBreakerEventListener.attachListeners(circuitBreaker);
```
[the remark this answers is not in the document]
pavan-traceable: No, its taken care using set now.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfigProvider.java
```diff
@@ -0,0 +1,86 @@
...
+
+  private void initialize(Config circuitBreakerConfig) {
+    circuitBreakerEnabled =
+        circuitBreakerConfig.hasPath(ENABLED) && circuitBreakerConfig.getBoolean(ENABLED);
+    this.circuitBreakerConfigurationMap =
+        circuitBreakerConfig.root().keySet().stream()
+            .filter(key -> !key.equals(ENABLED)) // Ignore the global enabled flag
+            .collect(
+                Collectors.toMap(
+                    key -> key, // Circuit breaker key
```
aaron-steinfeld, line 55: It seems like you're imagining all circuit breakers are siblings within one map - What's the benefit of that? IMO we should define some default json format that we can parse into a config, or let callers build their own programatically. Identical to how guava cache config works actually (although we usually use a custom format and programmatically translate it, the point is you can do it either way - this parser is just a convenience for uniformity).

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfigProvider.java
```diff
@@ -0,0 +1,86 @@
...
+        circuitBreakerConfigurationMap);
+  }
+
+  public Map<String, CircuitBreakerConfiguration> getConfigMap() {
+    return circuitBreakerConfigurationMap;
+  }
+
+  private CircuitBreakerConfiguration createCircuitBreakerConfig(Config config) {
+    return CircuitBreakerConfiguration.builder()
+        .failureRateThreshold((float) config.getDouble(FAILURE_RATE_THRESHOLD))
```
aaron-steinfeld, line 70: Does every field need to be defined in the config, or can we get away with defaults (or does the lib already do that for us?)

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfiguration.java
```diff
@@ -0,0 +1,33 @@
+package org.hypertrace.circuitbreaker.grpcutils;
+
+import java.time.Duration;
+import lombok.Builder;
+import lombok.Setter;
+import lombok.Value;
+
+@Value
+@Builder
+@Setter
```
aaron-steinfeld, line 10: nit - shouldn't be a setter _and_ value - value is immutable, builder allows setting.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerConfigParser.java
```diff
@@ -0,0 +1,41 @@
...
+/** Utility class to parse CircuitBreakerConfiguration to Resilience4j CircuitBreakerConfig */
+public class ResilienceCircuitBreakerConfigParser {
+
+  public static Map<String, CircuitBreakerConfig> getCircuitBreakerConfigs(
+      Map<String, CircuitBreakerConfiguration> configurationMap) {
+    return configurationMap.entrySet().stream()
+        .collect(Collectors.toMap(Map.Entry::getKey, entry -> getConfig(entry.getValue())));
+  }
+
+  static CircuitBreakerConfig getConfig(CircuitBreakerConfiguration configuration) {
```
aaron-steinfeld, line 17: nit: I would call the earlier class (that takes in config) more of a parser. This is more of a converter since both sides are structured + typed.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
```diff
@@ -0,0 +1,137 @@
...
+
+  public static final CallOptions.Key<String> CIRCUIT_BREAKER_KEY =
+      CallOptions.Key.createWithDefault("circuitBreakerKey", "default");
+  private final CircuitBreakerRegistry resilicenceCircuitBreakerRegistry;
+  private final CircuitBreakerConfigProvider circuitBreakerConfigProvider;
+  private final Map<String, CircuitBreakerConfig> resilienceCircuitBreakerConfig;
+  private final ResilienceCircuitBreakerProvider resilienceCircuitBreakerProvider;
+  private final Clock clock;
+
+  public ResilienceCircuitBreakerInterceptor(Config config, Clock clock) {
```
aaron-steinfeld, line 38: The interceptor should take the parsed config rather than doing it internally. We want to leave it up to the circuit breaker user to hand us a config so they can, for example, replace it in tests.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
```diff
@@ -0,0 +1,137 @@
...
+  @Override
+  protected boolean isCircuitBreakerEnabled() {
+    return circuitBreakerConfigProvider.isCircuitBreakerEnabled();
+  }
+
+  @Override
+  protected <ReqT, RespT> ClientCall<ReqT, RespT> createInterceptedCall(
+      MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+    // Get circuit breaker key from CallOptions
+    String circuitBreakerKey = callOptions.getOption(CIRCUIT_BREAKER_KEY);
```
aaron-steinfeld, line 76: Let's discuss this. I think using a global interceptor with magic attributes on the call is way more confusing and easy to mess up than a contract like below, where we attach the circuit breaker for our stub to our stub.


```
CircuitBreakerConfig parsedConfig = CircuitBreakerConfigParser.parse(genericConfig); // Or use the builder and roll your own
RequestInterceptor interceptor = new CircuitBreakerRequestInterceptor(parsedConfig, clock);

myStub = myStub.withInterceptor(interceptor);
// Now all calls via myStub go through the circuit breaker
```

grpc-circuitbreaker-utils/build.gradle.kts
```diff
@@ -0,0 +1,29 @@
...
+dependencies {
+
+  api(platform("io.grpc:grpc-bom:1.68.3"))
+  api("io.grpc:grpc-api")
+  api(project(":grpc-context-utils"))
+
+  implementation("org.slf4j:slf4j-api:1.7.36")
+  implementation("io.github.resilience4j:resilience4j-circuitbreaker:1.7.1")
+  implementation("com.typesafe:config:1.4.2")
+  implementation("com.google.inject:guice:7.0.0")
```
aaron-steinfeld, line 17: guice isn't used, can remove

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfiguration.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.Value;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+
+@Value
+@Builder
+public class CircuitBreakerConfiguration<T> {
+  Class<T> requestClass;
+  BiFunction<RequestContext, T, String> keyFunction;
+  boolean enabled;
+  Map<String, CircuitBreakerThresholds> circuitBreakerThresholdsMap;
```
aaron-steinfeld, line 15: We should probably have a `CircuitBreakerThresholds` - the map being just for any custom overrides (which would be supported but atypical).

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfigParser.java
```diff
@@ -0,0 +1,64 @@
...
+    CircuitBreakerConfiguration.CircuitBreakerConfigurationBuilder<T> builder =
+        CircuitBreakerConfiguration.builder();
+    Map<String, CircuitBreakerThresholds> circuitBreakerThresholdsMap =
+        config.root().keySet().stream()
+            .collect(
+                Collectors.toMap(
+                    key -> key, // Circuit breaker key
+                    key -> buildCircuitBreakerThresholds(config.getConfig(key))));
+    builder.circuitBreakerThresholdsMap(circuitBreakerThresholdsMap);
+    log.info("Loaded circuit breaker configs: {}", circuitBreakerThresholdsMap);
```
aaron-steinfeld, line 42: debug

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfigParser.java
```diff
@@ -0,0 +1,64 @@
...
+    return builder;
+  }
+
+  private static CircuitBreakerThresholds buildCircuitBreakerThresholds(Config config) {
+    return CircuitBreakerThresholds.builder()
+        .failureRateThreshold((float) config.getDouble(FAILURE_RATE_THRESHOLD))
```
aaron-steinfeld, line 48: Let's support defaults for all (or almost all) fields here - we shouldn't need to set every single field for this to work.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerConfigParser.java
```diff
@@ -0,0 +1,44 @@
...
+/** Utility class to parse CircuitBreakerConfiguration to Resilience4j CircuitBreakerConfig */
+public class ResilienceCircuitBreakerConfigParser {
+
+  public static Map<String, CircuitBreakerConfig> getCircuitBreakerConfigs(
+      Map<String, CircuitBreakerThresholds> configurationMap) {
+    return configurationMap.entrySet().stream()
+        .collect(Collectors.toMap(Map.Entry::getKey, entry -> getConfig(entry.getValue())));
+  }
+
+  static CircuitBreakerConfig getConfig(CircuitBreakerThresholds configuration) {
```
aaron-steinfeld, line 17: Naming - Since the config has already ben parsed, this is more of a converter or factory. Parsing happens with unstructured data.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
```diff
@@ -0,0 +1,147 @@
...
+    this.resilicenceCircuitBreakerRegistry =
+        new ResilienceCircuitBreakerRegistryProvider(resilienceCircuitBreakerConfigMap)
+            .getCircuitBreakerRegistry();
+    this.resilienceCircuitBreakerProvider =
+        new ResilienceCircuitBreakerProvider(
+            resilicenceCircuitBreakerRegistry, resilienceCircuitBreakerConfigMap);
+  }
+
+  @VisibleForTesting
+  public ResilienceCircuitBreakerInterceptor(
```
aaron-steinfeld, line 50: Visible for testing is an antipattern for testing old code - do we really need to test this with a separate registry? It's either an internal impl detail and we don't need to test it separately, or it should be passed in for the regular construction path.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
```diff
@@ -0,0 +1,147 @@
...
+      }
+
+      @SuppressWarnings("unchecked")
+      @Override
+      public void sendMessage(ReqT message) {
+        CircuitBreakerConfiguration<ReqT> config =
+            (CircuitBreakerConfiguration<ReqT>) circuitBreakerConfiguration;
+        if (config.getRequestClass() == null || config.getKeyFunction() == null) {
+          log.debug("Circuit breaker will apply to all requests as config is not set");
+          circuitBreakerKey = "default";
```
aaron-steinfeld, line 93: This is an assumption we don't want to make - no magic keys. Let's make the default built into the config.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
```diff
@@ -0,0 +1,147 @@
...
+        } else {
+          if (!message.getClass().equals(config.getRequestClass())) {
```
aaron-steinfeld, line 95: Can we move this out of the interceptor and just pass the message to the config and ask for the response? inside, it can pull the class of the message if needed and map the key.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
```diff
@@ -0,0 +1,147 @@
...
+            log.warn("Invalid config for message type: {}", message.getClass());
```
aaron-steinfeld, line 96: Not necessarily an error. We could use a function that's not specific to a class (e.g. by tenant id).

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerRegistryProvider.java
```diff
@@ -0,0 +1,40 @@
...
+  public CircuitBreakerRegistry getCircuitBreakerRegistry() {
+    CircuitBreakerRegistry circuitBreakerRegistry =
+        CircuitBreakerRegistry.of(
+            this.circuitBreakerConfigMap.get(CircuitBreakerConfigParser.DEFAULT_CONFIG_KEY));
+    circuitBreakerRegistry
+        .getEventPublisher()
+        .onEntryAdded(
+            entryAddedEvent -> {
+              CircuitBreaker addedCircuitBreaker = entryAddedEvent.getAddedEntry();
+              log.info("CircuitBreaker {} added", addedCircuitBreaker.getName());
```
aaron-steinfeld, line 31: We really abuse info logs here - can this be debug? Or will it only happen once at startup?

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
pavan-traceable: We need to mock these classes for testing. One alternative way is creating factory and client code will look like 
`new ResilienceCircuitBreakerInterceptor(
    config,
    clock,
    CircuitBreakerDependencyFactory.createRegistry(config.getDefaultThresholds()),
    CircuitBreakerDependencyFactory.createProvider(
        registry, 
        ResilienceCircuitBreakerConfigConverter.getCircuitBreakerConfigs(config.getCircuitBreakerThresholdsMap())
    )
);
`

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
pavan-traceable: Discussed offline and updated the patch

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
```diff
@@ -0,0 +1,142 @@
...
+      @SuppressWarnings("unchecked")
+      @Override
+      public void sendMessage(ReqT message) {
+        CircuitBreakerConfiguration<ReqT> config =
+            (CircuitBreakerConfiguration<ReqT>) circuitBreakerConfiguration;
+        if (config.getKeyFunction() == null) {
+          log.debug("Circuit breaker will apply to all requests as keyFunction config is not set");
+          circuitBreakerKey = config.getDefaultCircuitBreakerKey();
+        } else {
+          circuitBreakerKey = config.getKeyFunction().apply(RequestContext.CURRENT.get(), message);
```
aaron-steinfeld, line 94: We should support there being no default too - e.g. I can specify a default or not, overrides or not. If no key and no default, then it would not skip any circuit breaker.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+        circuitBreaker = resilienceCircuitBreakerProvider.getCircuitBreaker(circuitBreakerKey);
+        if (!circuitBreaker.tryAcquirePermission()) {
+          logCircuitBreakerRejection(circuitBreakerKey, circuitBreaker);
+          String rejectionReason =
+              circuitBreaker.getState() == CircuitBreaker.State.HALF_OPEN
+                  ? "Circuit Breaker is HALF-OPEN and rejecting excess requests"
+                  : "Circuit Breaker is OPEN and blocking requests";
+          throw Status.RESOURCE_EXHAUSTED.withDescription(rejectionReason).asRuntimeException();
```
aaron-steinfeld, line 103: So a nice-to-have would be allowing the config to provide an exception builder with this being the default.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfigParser.java
aaron-steinfeld: Because the config is already logged at service startup, so this is duplicative. In general, libraries shouldn't do any non-debug logging without configuration options since consumers can't opt out - so either we add in config to control these (or hooks for clients to inject their own) or make them debug.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
aaron-steinfeld: We may also want to check that the message matches the class if there is a class specified - either doing that here, or wrapping the config function inside the config works too. If we don't, we're requiring the caller to do it without type safety (because we're using the cast on line 89 to make the assumption they already did) and if they don't they'll get a runtime error.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
pavan-traceable: Agree, to mitigate the behaviour of different scenarios, we need to mock registry and provider. But is it ok in the client side we need to pass `CircuitBreakerDependencyFactory.createRegistry(config.getDefaultThresholds()), CircuitBreakerDependencyFactory.createProvider( registry,  ResilienceCircuitBreakerConfigConverter.getCircuitBreakerConfigs(config.getCircuitBreakerThresholdsMap())  as constructor arguments` these as well. Won't these unnecessary to consumer ?

pavan-traceable: Added a factory and fixed it.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
pavan-traceable: Even without defaults provided from consumer it will still take defaults and it will work.

pavan-traceable: Added validation for Request class mismatch.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
```diff
@@ -0,0 +1,59 @@
...
+  private final Map<String, CircuitBreaker> circuitBreakerCache = new ConcurrentHashMap<>();
+
+  public ResilienceCircuitBreakerProvider(
+      CircuitBreakerRegistry circuitBreakerRegistry,
+      Map<String, CircuitBreakerConfig> circuitBreakerConfigMap) {
+    this.circuitBreakerRegistry = circuitBreakerRegistry;
+    this.circuitBreakerConfigMap = circuitBreakerConfigMap;
+  }
+
+  public CircuitBreaker getCircuitBreaker(String circuitBreakerKey) {
```
aaron-steinfeld, line 26: This should return optional. We want to find the one that matches the key, if it exists; else the default, if it exists. Else the call should not go through a circuit breaker.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
```diff
@@ -0,0 +1,124 @@
...
+        super.start(wrappedListener, headers);
+      }
+
+      @SuppressWarnings("unchecked")
+      @Override
+      public void sendMessage(ReqT message) {
+        CircuitBreakerConfiguration<ReqT> config =
+            (CircuitBreakerConfiguration<ReqT>) circuitBreakerConfiguration;
+        // Type check for message class compatibility
+        if (config.getRequestClass() != null && !config.getRequestClass().isInstance(message)) {
```
aaron-steinfeld, line 66: This doesn't need to hold. Treat the request class as a filter. If I add an interceptor to a stub for one method of the stub, the other methods should not be circuit broken. Since we can't add the calls more granularly than the stub level, we'll never have interceptors that only are being called for one specific class.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfigParser.java
pavan-traceable: Fixed it

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfigParser.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class CircuitBreakerConfigParser {
+
+  private static final Set<String> nonThresholdKeys =
+      Set.of("enabled", "defaultCircuitBreakerKey", "defaultThresholds");
```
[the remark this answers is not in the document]
pavan-traceable: done

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfiguration.java
```diff
@@ -0,0 +1,28 @@
...
+import org.hypertrace.core.grpcutils.context.RequestContext;
+
+@Value
+@Builder
+public class CircuitBreakerConfiguration<T> {
+  Class<T> requestClass;
+  BiFunction<RequestContext, T, String> keyFunction;
+  @Builder.Default boolean enabled = false;
+  // Default value be "global" if not override.
+  String defaultCircuitBreakerKey = "global";
```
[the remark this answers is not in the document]
pavan-traceable: It is getting used here 
`if (config.getKeyFunction() == null) {
          log.debug("Circuit breaker will apply to all requests as keyFunction config is not set");
          circuitBreakerKey = config.getDefaultCircuitBreakerKey();
        } `

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
pavan-traceable: We are maintaining a cache here, if circuitBreaker key exists it is not going through circuit breaker

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
pavan-traceable: Couldn't understand this, what need to be done here ?

aaron-steinfeld: Responded offline, but basically this request class is a criteria to be met before we assign a circuit breaker. If it's not met, we should skip the circuit breaker rather than erroring. This is because we attach interceptors at the stub or channel level, so if the circuit breaker is for a single method we want it to ignore all other methods. If it's for multiple methods, it shouldn't be keyed by the request (although we could support multiple methods keying off an aspect of the context like tenant).

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
pavan-traceable: Not able to understand here, i have setup meeting to discuss and close this.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
pavan-traceable: Fixed it

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfigParser.java
```diff
@@ -38,20 +38,19 @@ public static <T> CircuitBreakerConfiguration.CircuitBreakerConfigurationBuilder
...
         config.root().keySet().stream()
             .filter(key -> !NON_THRESHOLD_KEYS.contains(key)) // Filter out non-threshold keys
             .collect(
                 Collectors.toMap(
                     key -> key, // Circuit breaker key
                     key -> buildCircuitBreakerThresholds(config.getConfig(key))));
+
+    if (!config.hasPath(DEFAULT_THRESHOLDS)) {
+      builder.defaultThresholds(buildCircuitBreakerDefaultThresholds());
+      circuitBreakerThresholdsMap.put(DEFAULT_THRESHOLDS, buildCircuitBreakerDefaultThresholds());
```
aaron-steinfeld, line 51: Why are we writing this into the map? The the default thresholds don't need to be accessed by key - that should be what `builder.defaultThresholds` is for. Also note if `DEFAULT_THRESHOLDS` path _does_ exist it would make it into the map, but it would not be set to `builder.defaultThresholds`. I would expect this logic to look more like
```java
builder.defaultThresholds(
  config.hasPath(DEFAULT_THRESHOLDS) ?
    buildCircuitBreakerThresholds(config.getConfig(DEFAULT_THRESHOLDS)) : 
    buildCircuitBreakerDefaultThresholds());
```

The config key then is not a concern of the specific impl. If you really want to use a magic key, then `defaultThresholds` would be removed, we wouldn't want both since it's two ways of conveying the same thing - one just decouples concerns.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerFactory.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class ResilienceCircuitBreakerFactory {
+  public static ResilienceCircuitBreakerInterceptor getResilienceCircuitBreakerInterceptor(
+      CircuitBreakerConfiguration<?> circuitBreakerConfiguration, Clock clock) {
+    Map<String, CircuitBreakerConfig> resilienceCircuitBreakerConfigMap =
+        ResilienceCircuitBreakerConfigConverter.getCircuitBreakerConfigs(
+            circuitBreakerConfiguration.getCircuitBreakerThresholdsMap());
+    CircuitBreakerRegistry resilicenceCircuitBreakerRegistry =
+        new ResilienceCircuitBreakerRegistryProvider(
+                circuitBreakerConfiguration.getDefaultThresholds())
```
aaron-steinfeld, line 17: So these default thresholds go into the registry directly, but the others are built as-needed?

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerFactory.java
```diff
@@ -18,7 +18,9 @@ public static ResilienceCircuitBreakerInterceptor getResilienceCircuitBreakerInt
             .getCircuitBreakerRegistry();
     ResilienceCircuitBreakerProvider resilienceCircuitBreakerProvider =
         new ResilienceCircuitBreakerProvider(
-            resilicenceCircuitBreakerRegistry, resilienceCircuitBreakerConfigMap);
+            resilicenceCircuitBreakerRegistry,
+            resilienceCircuitBreakerConfigMap,
+            circuitBreakerConfiguration.getCircuitBreakerThresholdsMap());
```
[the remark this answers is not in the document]
pavan-traceable: done.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerFactory.java
```diff
@@ -18,7 +18,9 @@ public static ResilienceCircuitBreakerInterceptor getResilienceCircuitBreakerInt
             .getCircuitBreakerRegistry();
     ResilienceCircuitBreakerProvider resilienceCircuitBreakerProvider =
         new ResilienceCircuitBreakerProvider(
-            resilicenceCircuitBreakerRegistry, resilienceCircuitBreakerConfigMap);
+            resilicenceCircuitBreakerRegistry,
```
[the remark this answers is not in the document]
pavan-traceable: done

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
```diff
@@ -67,17 +68,21 @@ public void sendMessage(ReqT message) {
           super.sendMessage(message);
           return;
         }
-        if (config.getKeyFunction() == null) {
-          log.debug("Circuit breaker will apply to all requests as keyFunction config is not set");
-          circuitBreakerKey = config.getDefaultCircuitBreakerKey();
-        } else {
+        if (config.getKeyFunction() != null) {
           circuitBreakerKey = config.getKeyFunction().apply(RequestContext.CURRENT.get(), message);
+          circuitBreaker = resilienceCircuitBreakerProvider.getCircuitBreaker(circuitBreakerKey);
```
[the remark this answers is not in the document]
pavan-traceable: We are setting default always why will circuitBreaker become null ?
 builder.defaultThresholds(
  config.hasPath(DEFAULT_THRESHOLDS) ?
    buildCircuitBreakerThresholds(config.getConfig(DEFAULT_THRESHOLDS)) : 
    buildCircuitBreakerDefaultThresholds());

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfigParser.java
pavan-traceable: done

pavan-traceable: This is done mainly to handle where circuit breaker is disabled by default but enabled for only tenant, in CircuitBreakerProvider we check enabled or based on CircuitConfigMap only, so thats why adding to map.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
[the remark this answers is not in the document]
pavan-traceable: Why keyFunction should be null ? To handle what scenarios ? Can you please list all scenarios that we need to handle, i will check accordingly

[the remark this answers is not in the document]
aaron-steinfeld: > We are setting default always why will circuitBreaker become null ?

We're requiring config for it, but it can be disabled and would not have an associated circuit breaker.

> If someone defines keyFunction, why key should be null ? 

Imagine an API like "doAction(action)" where we want to circuit break some expensive action like "email" by tenant ID. A key function might look like `action == "email" ? getTenantId() : null`. That would allow other actions to continue on without a circuit breaker, but emails to be circuit broken by tenant ID. Alternatively, we could add a separate optional filter method with a contract like (context, request) => boolean. Do you think that would be clearer?

I realize looking back my logic was a bit off here too.

> Can you please list all scenarios that we need to handle, i will check accordingly

Basically any permutation - should be able to provide a key function (and filter/predicate, if we want to separate that - the class match is a bit of an implicit predicate but does not encompass all scenarios) which allows us to:
 
1. route some requests to custom circuit breaker that's declared in config
2. route others to a default circuit breaker that's declared in config
3. have others skip the circuit breaker entirely

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
```diff
@@ -0,0 +1,73 @@
...
+  public ResilienceCircuitBreakerProvider(
+      CircuitBreakerRegistry circuitBreakerRegistry,
+      Map<String, CircuitBreakerConfig> circuitBreakerConfigMap,
+      List<String> disabledKeys) {
+    this.circuitBreakerRegistry = circuitBreakerRegistry;
+    this.circuitBreakerConfigMap = circuitBreakerConfigMap;
+    this.disabledKeys = disabledKeys;
+  }
+
+  public Optional<CircuitBreaker> getCircuitBreaker(String circuitBreakerKey) {
```
aaron-steinfeld, line 31: So if this is some random key (that would map to the default) how do we handle there being no default here? It looks like the disabled key is only being respected for the overrides

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
```diff
@@ -0,0 +1,73 @@
...
+                          log.debug(
+                              "Circuit breaker event type {} for circuit breaker name {}",
+                              event.getEventType(),
+                              event.getCircuitBreakerName()));
+              return circuitBreaker;
+            }));
+  }
+
+  public Optional<CircuitBreaker> getDefaultCircuitBreaker() {
+    return getCircuitBreaker(CircuitBreakerConfigParser.DEFAULT_THRESHOLDS);
```
aaron-steinfeld, line 65: So I see one part of the confusion here. You're using this special key for the "shared" circuit breaker that's used when there is no key. We can use a special key for that, but it should be local to this provider - it's indepdendent of the config.

I think part of the confusion is that the term default is being overloaded a bit here. There's two separate things:
- the config to use if no key matches (which can be disabled - if no specific key config, don't circuit break)
- the shared circuit breaker to use if there is no key (which would be an instance of the former case - the config to use if no key matches).

pavan-traceable: Thanks for your patience, i hope i have addressed all the cases now. Due to multiple scenarios, i feel implementation is slightly complicated, not sure if we use these different scenarios.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
pavan-traceable: Fixed it.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerInterceptor.java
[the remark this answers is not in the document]
pavan-traceable: Hopefully i have fixed all different scenarios now. Please check

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfigParser.java
```diff
@@ -0,0 +1,111 @@
...
+      Config config) {
+    CircuitBreakerConfiguration.CircuitBreakerConfigurationBuilder<T> builder =
+        CircuitBreakerConfiguration.builder();
+    if (config.hasPath(ENABLED)) {
+      builder.enabled(config.getBoolean(ENABLED));
+    }
+
+    Map<String, CircuitBreakerThresholds> circuitBreakerThresholdsMap =
+        config.root().keySet().stream()
+            .filter(key -> !NON_THRESHOLD_KEYS.contains(key)) // Filter out non-threshold keys
```
aaron-steinfeld, line 43: Because NON_THRESHOLD_KEYS is not excluding the default, the default will be put into the map like every other config in addition to being set as `defaultThresholds`.

Can avoid putting it in the map any only use it through `defaultThresholds`? Because also note that if I choose to build the config in code (rather than parsing it from config) I would only set it in the one place.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
```diff
@@ -0,0 +1,99 @@
...
+  private final boolean defaultEnabled;
+
+  public ResilienceCircuitBreakerProvider(
+      CircuitBreakerRegistry circuitBreakerRegistry,
+      Map<String, CircuitBreakerConfig> circuitBreakerConfigMap,
+      List<String> disabledKeys) {
+    this.circuitBreakerRegistry = circuitBreakerRegistry;
+    this.circuitBreakerConfigMap = circuitBreakerConfigMap;
+    this.disabledKeys = disabledKeys;
+    this.defaultEnabled = !disabledKeys.contains(CircuitBreakerConfigParser.DEFAULT_THRESHOLDS);
```
aaron-steinfeld, line 31: This works because the defaults are still being put in the map along with every other config - but we can't rely on that as I pointed out in an earlier comment (it wouldn't hold if I built my config rather than parsed it, and it's that same abstraction leak we've been trying to avoid).

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
```diff
@@ -0,0 +1,99 @@
...
+              attachListeners(circuitBreaker);
+              return circuitBreaker;
+            }));
+  }
+
+  public Optional<CircuitBreaker> getSharedCircuitBreaker() {
+    if (!defaultEnabled) {
+      return Optional.empty();
+    }
+    return Optional.of(
```
aaron-steinfeld, line 57: nit - can this just be `return getCircuitBreaker(SHARED_KEY);`?

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
```diff
@@ -0,0 +1,99 @@
...
+
+  private CircuitBreaker getCircuitBreakerFromConfigMap(
+      String circuitBreakerKey, boolean defaultEnabled) {
+    return Optional.ofNullable(circuitBreakerConfigMap.get(circuitBreakerKey))
+        .map(config -> circuitBreakerRegistry.circuitBreaker(circuitBreakerKey, config))
+        .orElseGet(
+            () ->
+                defaultEnabled
+                    ? circuitBreakerRegistry.circuitBreaker(circuitBreakerKey)
+                    : null); // Return null if default is disabled
```
aaron-steinfeld, line 97: Nit - given that most of this PR (including this same function) is working in optionals, please stick to optionals over null unless the third party code doesn't give us an option. This function can return an optional to convey it like
```java
private Optional<CircuitBreaker> buildNewCircuitBreaker(String circuitBreakerKey, boolean defaultEnabled) {
  return Optional.ofNullable(circuitBreakerConfigMap.get(circuitBreakerKey))
        .map(config -> circuitBreakerRegistry.circuitBreaker(circuitBreakerKey, config))
        .or(() -> defaultEnabled ? Optional.of(circuitBreakerRegistry.circuitBreaker(circuitBreakerKey)) : Optional.empty());
```

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+    return Optional.ofNullable(
+        circuitBreakerCache.computeIfAbsent(
+            circuitBreakerKey,
+            key -> {
+              CircuitBreaker circuitBreaker =
+                  getCircuitBreakerFromConfigMap(circuitBreakerKey, defaultEnabled);
+              // If no circuit breaker is created return empty
+              if (circuitBreaker == null) {
+                return null; // Ensures cache does not store null entries
```
aaron-steinfeld, line 46: Why not store nulls? No need to recompute the disabled status for the same key. Could make the map value an optional to avoid the null behavior of computeIfAbsent.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
```diff
@@ -0,0 +1,99 @@
...
+import org.hypertrace.circuitbreaker.grpcutils.CircuitBreakerConfigParser;
+
+/** Utility class to provide Resilience4j CircuitBreaker */
+@Slf4j
+class ResilienceCircuitBreakerProvider {
+
+  private static final String SHARED_KEY = "SHARED_KEY";
+  private final CircuitBreakerRegistry circuitBreakerRegistry;
+  private final Map<String, CircuitBreakerConfig> circuitBreakerConfigMap;
+  private final Map<String, CircuitBreaker> circuitBreakerCache = new ConcurrentHashMap<>();
```
aaron-steinfeld, line 20: Are there any concerns about this growing too large? If we were to use a bad key (e.g. request ID) as our circuit breaker it would grow unbounded. Consider if we should use a true simple cache here like below to prevent a bad config from crashing a service.
```
        CacheBuilder.newBuilder()
            .expireAfterAccess(Duration.ofHours(1)) // or from config
            .maximumSize(10_000)
            .build(CacheLoader.from(this::buildCircuitBreaker));
```

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/CircuitBreakerConfigParser.java
pavan-traceable: Fixed it

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
pavan-traceable: fixed it

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
pavan-traceable: resolved

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
pavan-traceable: My bad, updated

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
pavan-traceable: Updated

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
pavan-traceable: fixed it

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
```diff
@@ -17,51 +19,42 @@ class ResilienceCircuitBreakerProvider {
...
+      CacheBuilder.newBuilder()
+          .expireAfterAccess(60, TimeUnit.MINUTES) // Auto-evict after 60 minutes
+          .maximumSize(10000) // Limit max cache size
+          .build(
+              new CacheLoader<>() {
+                @Override
+                public Optional<CircuitBreaker> load(String key) {
+                  return buildNewCircuitBreaker(key);
+                }
+              });
```
aaron-steinfeld, line 36: nit
```suggestion
          .build(CacheLoader.from(this::buildNewCircuitBreaker));
```

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
```diff
@@ -86,14 +79,26 @@ private static void attachListeners(CircuitBreaker circuitBreaker) {
...
+              if (defaultEnabled) {
+                CircuitBreaker circuitBreaker =
+                    circuitBreakerRegistry.circuitBreaker(circuitBreakerKey);
+                attachListeners(
+                    circuitBreaker); // Attach listeners here for default circuit breaker
+                return Optional.of(circuitBreaker);
+              } else {
+                return Optional.empty();
+              }
+            });
```
aaron-steinfeld, line 102: nit - can do shared things at the end once you've built the circuit breaker
```java
     return Optional.ofNullable(circuitBreakerConfigMap.get(circuitBreakerKey))
        .map(config -> circuitBreakerRegistry.circuitBreaker(circuitBreakerKey, config))
        .or(() -> {
              if (defaultEnabled) {
                return Optional.of(circuitBreakerRegistry.circuitBreaker(circuitBreakerKey));
              } else {
                return Optional.empty();
              }
            })
        .map(this::attachListeners); // assuming attach listeners can be made fluent and return the circuitbreaker
```

aaron-steinfeld (approved):
Thanks for working through all the comments on this one - this will be a nice add!
Left a couple minor nits that are just style prefs, but otherwise approved.

grpc-circuitbreaker-utils/src/main/java/org/hypertrace/circuitbreaker/grpcutils/resilience/ResilienceCircuitBreakerProvider.java
pavan-traceable: Will fix in immediate PR, merging this

pavan-traceable: https://github.com/hypertrace/java-grpc-utils/pull/69/files Please review

PR #70, opened 2025-04-08
Grpc cb utility update

Gg-harness:
## Description
Added config key support to define nested (channel level) configurations.

For Ex : 
Circuit Breaker Key - TenantId (defined by key function)
Config Key : S3 (notification channel)
Even though circuit breaker objects can be different for different tenants, we can define the CB configuration at channel level

### Testing
Tested changes on sandbox environment using this utility. Unit Tests added and validated.
Scenarios : 
1. Thresholds defined for circuit breaker key => should pick thresholds against circuit breaker key 
2. Thresholds not defined for circuit breaker key
	i. Config Key Set, Thresholds defined for config key => should pick thresholds against config key
	ii. Config Key Set, Thresholds not defined for config key => should pick default thresholds
	iii. Config key not set => should pick default thresholds
	
Jira : https://harness.atlassian.net/browse/ENGTAI-60244


aaron-steinfeld (changes requested):
Discussing offline - use multiple circuit breakers rather than adding extra levels of scoping inside the config.

PR #93, opened 2026-04-24
ASP-2043: Add typed gRPC built-in retry support to GrpcChannelConfig

kunalvaswani123:
## Summary

- Add `GrpcRetryPolicy` — typed retry config (maxAttempts, backoff, retryable status codes)
- Add `GrpcServiceConfig` — wraps `GrpcRetryPolicy`, owns conversion to gRPC service config map
- Add `serviceConfig: GrpcServiceConfig` and `enableRetry: boolean` to `GrpcChannelConfig` as independent opt-ins
- Wire both in `GrpcChannelRegistry.configureAndBuildChannel()` — no retry assumed from service config alone

## Test plan
- `GrpcChannelRegistryRetryTest` — starts a real in-process gRPC server that fails twice with `UNAVAILABLE`, succeeds on the 3rd attempt, and asserts the server was hit 3 times from a single client call

🤖 Generated with [Claude Code](https://claude.com/claude-code)


grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
```diff
@@ -90,6 +90,9 @@ protected ManagedChannel configureAndBuildChannel(
     if (config.getMaxInboundMessageSize() != null) {
       builder.maxInboundMessageSize(config.getMaxInboundMessageSize());
     }
+    if (config.getServiceConfig() != null) {
+      builder.defaultServiceConfig(config.getServiceConfig()).enableRetry();
```
aaron-steinfeld: We would want to type this, if exposing. that's the purpose and benefit of the abstraction.

And simply setting a config does not imply enabling retry either

aaron-steinfeld (changes requested):

grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
kunalvaswani123: @aaron-steinfeld  added GrpcServiceConfig, GrpcRetryPolicy types
Also, I've added enableRetry to enable the retries

grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcRetryPolicy.java
```diff
@@ -0,0 +1,41 @@
...
+  Duration maxBackoff;
+  double backoffMultiplier;
+  @Singular List<Status.Code> retryableStatusCodes;
+
+  Map<String, Object> toMap() {
+    return Map.of(
+        MAX_ATTEMPTS,
+        (double) maxAttempts,
+        INITIAL_BACKOFF,
+        initialBackoff.toMillis() / 1000.0 + "s",
```
aaron-steinfeld: What's the format here? If it's a proto Duration (as declared in the proto) then convert it and toString it. Avoid hand-conversions.

grpc-client-utils/src/test/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistryTest.java
```diff
@@ -145,6 +147,31 @@ void copyConstructorReusesExistingChannels() {
...
+                    .retryableStatusCode(Status.Code.UNAVAILABLE)
+                    .build())
+            .build();
+    Channel channelWithServiceConfig =
+        this.channelRegistry.forSecureAddress(
+            "foo", 1000, GrpcChannelConfig.builder().serviceConfig(serviceConfig).build());
+
+    assertNotNull(channelWithServiceConfig);
+    assertNotSame(channelWithServiceConfig, this.channelRegistry.forSecureAddress("foo", 1000));
+    assertSame(
```
aaron-steinfeld: Can we add a test that actual sees if this config applies? Do we have any in process tests in this repo?

grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
```diff
@@ -90,6 +90,9 @@ protected ManagedChannel configureAndBuildChannel(
     if (config.getMaxInboundMessageSize() != null) {
       builder.maxInboundMessageSize(config.getMaxInboundMessageSize());
     }
+    if (config.getServiceConfig() != null) {
+      builder.defaultServiceConfig(config.getServiceConfig().toMap()).enableRetry();
```
aaron-steinfeld: move enable retry to top level as an opt in. We don't want to assume that providing a service config (Which may be entirely unrelated to retrying) implies the caller wants retry.

grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcRetryPolicy.java
kunalvaswani123: using com.google.protobuf.util.Durations now for conversion

grpc-client-utils/src/main/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistry.java
kunalvaswani123: Added a boolean for this in GrpcChannelConfig

grpc-client-utils/src/test/java/org/hypertrace/core/grpcutils/client/GrpcChannelRegistryTest.java
kunalvaswani123: Added GrpcChannelRegistryRetryTest with an in-process test (retriesOnUnavailableWhenRetryPolicyConfigured) that verifies the retry config

.gitignore
```diff
@@ -5,3 +5,5 @@
 build
 
 .idea
+
+settings-gradle.lockfile
```
aaron-steinfeld: note the reason we don't typically ignore this file is we don't want it on our local, either. If it's ignored it makes it easy to miss when it gets generated.

kunalvaswani123: removed the gitignore

aaron-steinfeld (approved):

